PR #64, opened 2018-11-03
Fixes formatting in Create individual

Yuri05:
Just another case where formatting looks good in MD but Gitbook destroys it.
Tried to change it - let's see if it's better now after rendering.


Yuri05:
Still corrupted after rendering.. ok, let's wait for Gitbook answer

msevestre:
Not sure what issue you are talking about? Looks fine to me

msevestre:
image

Yuri05:
That's how it looks like in MD and how it should be (s. scaling method options)

grafik

PR #84, opened 2019-11-02
83 page empty

msevestre:


msevestre:
@Yuri05 There was some weird invisible characters making the link not valid
Also the indentation was completely off, as well as some of the markers
Last I have renamed (re)-qualification to qualfiicaton because  the former does not mean anything to anyome except to a very few selected people

Yuri05:
Crazy...

msevestre:
@Yuri05 You could see the issue right from vscode. The link was not valid. (e.g. hovering over the link would say page-5/requalification.md not found.

PR #86, opened 2019-11-02
fix image size

Yuri05:


Yuri05:
No, we should not resize original images.
And I even wouldn't have any idea how to resize them.
The problem is GitBook-rendering. All images look good in the source-markdown (without any resizing).
However some of them are not properly rendered by GitBook. So I just tried some options (e.g. HTML resizing), which didn't help.

Yuri05:
Btw: any chance to integrate a GitBook-Preview into PRs?

PR #101, opened 2020-04-16
Update icons

msevestre:


msevestre:
@Yuri05 The idea is to only use icons from the ospsuite (*.ico) if possible. We have set the size to 32 everywhere to have a consistent display

msevestre:
@Yuri05 If you are ok with these changes, please merge
Next, we will move all screenshots that are under icons in under the image folders (so that we do not mix and match)

Then we found a bunch of inconsistencies that we need to discuss
anything else?

Yuri05:
@msevestre In the changed markdown some new icons are not shown due to case sensitivity(?)
E.g. because icons are referenced as ".../human.ico" instead of ".../Human.ico" it looks like this:
grafik
If later for the rendered website it does not matter - we can merge

msevestre:
no way for me to know if this will be ok or not. vscode has no problem with it
Are you sure this is due to casing?

msevestre:
ok let's make sure we use the right casing.

Yuri05:
>Are you sure this is due to casing?

Yes. E.g. enter in Firefox
https://github.com/Open-Systems-Pharmacology/docs/blob/update-icons/assets/icons/human.ico
vs.
https://github.com/Open-Systems-Pharmacology/docs/blob/update-icons/assets/icons/Human.ico

Which actually really surprises me

PR #118, opened 2020-12-28
Add CheckLinks/SpellCheck actions

Yuri05:


.github/workflows/MarkdownLinksCheck.yml
```diff
@@ -0,0 +1,19 @@
+name: Check Markdown links
```
Yuri05, line 1: This action checks external links (e.g. `http://tralala.com`) and links to relative files (e.g. `/images/tralala.png`) but not the internal links withing an md-document (e.g. `#5-references`)

.github/workflows/MarkdownLinksCheck.yml
```diff
@@ -0,0 +1,19 @@
...
+
+on: 
+  pull_request:
+    branches:
+      - master
```
Yuri05, line 6: execute action on pull requests into master

.github/workflows/MarkdownLinksCheck.yml
```diff
@@ -0,0 +1,19 @@
...
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        ref: ${{ github.event.pull_request.head.sha }}
+    - uses: gaurav-nelson/github-action-markdown-link-check@v1
+      with:
+        use-quiet-mode: 'yes'
+        use-verbose-mode: 'yes'
+        config-file: '.github/workflows/mlc_config.json'
```
Yuri05, line 19: additional configuration options are stored in mlc_config.json

.github/workflows/SpellChecker.yml
```diff
@@ -0,0 +1,15 @@
+name: Spellcheck
```
Yuri05, line 1: Spellcheck action based on dictionaries

.github/workflows/XRefCheck.yml
```diff
@@ -0,0 +1,15 @@
+name: XRefCheck
```
Yuri05, line 1: This action checks links to relative files (e.g. `/images/tralala.png`) and the internal links withing an md-document (e.g. `#5-references`). External links (e.g. `http://tralala.com`) are theoretically checked as well, but not as good as by another checker (some wrong external links are not shown)

.github/workflows/mlc_config.json
```diff
@@ -0,0 +1,6 @@
+{
+  "retryOn429": true,
+  "retryCount": 5,
+  "fallbackRetryDelay": "30s",
+  "aliveStatusCodes": [200, 403]
+}
```
Yuri05, line 6: Additional configuration for the first link checker (s. https://github.com/gaurav-nelson/github-action-markdown-link-check for details)
It seems that some sites (e.g. github itself) block such script actions. In such a case, if the link is valid but the action cannot access it, 403 code is returned. So I added it to the list of allowed return codes (200 is: link is valid)

.spellcheck.yml
```diff
@@ -0,0 +1,18 @@
+matrix:
```
Yuri05, line 1: Additional configuration for the spell checker

.spellcheck.yml
```diff
@@ -0,0 +1,18 @@
...
+- name: Markdown
+  aspell:
+    lang: en
+  dictionary:
+    wordlists:
+    - .github/workflows/wordlist.txt
```
Yuri05, line 7: additional dictionary of allowed words

.github/workflows/MarkdownLinksCheck.yml
```diff
@@ -0,0 +1,19 @@
...
+      - master
+
+jobs:
+  markdown-link-check:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        ref: ${{ github.event.pull_request.head.sha }}
+    - uses: gaurav-nelson/github-action-markdown-link-check@2a60e0fe41b5361f446ccace6621a1a2a5c324cf
```
msevestre, line 15: Can't you use a tagged version instead? using a sha looks really weird

msevestre (commented):
This looks awesome.
Can we reference TAGGED version of script instead of SHA? This would be my only concern

.github/workflows/MarkdownLinksCheck.yml
Yuri05: using SHA is just more safe (e.g. the code in a tag could be - theoretically - replaced by the 3rd party vendor via removing the tag and creating the new tag with the same name and the new code)
Is actually what GitHub currently suggests for the 3rd-party actions
https://docs.github.com/en/free-pro-team@latest/actions/learn-github-actions/security-hardening-for-github-actions#using-third-party-actions
But I agree, with SHA you have to dig into the action-repo to figure out which vesion is currently used. Hope GitHub will enable one day using a Tag **together** with SHA or something like that.

msevestre: I think this is overkill for us, specifically considering that everything is opened source already and there is absolutely no secret info to gain that is not visible. Using a tag would just be sufficient in my opinion. 
And if an action would in the very very very unlikely even delete our content..so what. We can add it back. So I would still suggest using tags

PR #136, opened 2021-06-24
Importer redesign

msevestre:


msevestre:
@georgeDaskalakis  @Yuri05  MErging into v10

PR #155, opened 2021-09-21
edits to documentation website

IndrajeetPatil:
* fixes missing reference links
* fixes a little punctuation and spelling mistakes
* rephrasing of some awkwardly structured sentences
* no major changes to content


IndrajeetPatil:
@msevestre This is still a draft PR, but just wanted to point out that the GHA failure doesn't seem to have anything to do with the changes I made since I haven't touched any of the workflow files.

msevestre:
@IndrajeetPatil This is the first time that we have two versions so we are a bit behind on this. Eventually, we will have
v9
main == v10

and a v11 branch probably 
So yes, I agree, explaining what branch to use with a PR is an absolute must.

PR #160, opened 2021-09-30
156 update r version

msevestre:


msevestre:
@Yuri05 Merging to see how it looks like.

PR #181, opened 2022-03-29
175 ckd

msevestre:


references.md
```diff
@@ -1,3 +1,4 @@
+<!-- markdown-link-check-disable -->
```
Yuri05, line 1: @msevestre I don't think it's a good idea.
Better: extend the list of "is ok" status codes.
E.g. when I look on the messages from **this** link checker in the history: it seems to be always the 503 code.
So if you extend 
https://github.com/Open-Systems-Pharmacology/docs/blob/0b47c1e9de4480abd4e20cac577156c493664a23/.github/workflows/mlc_config.json#L5
with 503: should be fine (at least for **this** link checker) - and the links are still checked

.github/workflows/mlc_config.json
```diff
@@ -2,5 +2,5 @@
   "retryOn429": true,
   "retryCount": 5,
   "fallbackRetryDelay": "30s",
-  "aliveStatusCodes": [200, 403]
+  "aliveStatusCodes": [200, 403, 503]
```
Yuri05, line 5: seems to work :)
The one link checker is green now.
And obviously the GH actions are not optimally setup when working with **branches**. I always work with forks, so it's fine for me. But now it creates 1 job for push (in the originl branch) and 1 job for PR (in the target branch).

We could remove the `on push` condition and keep `on pull_request` only

msevestre: how do you do this?

PR #194, opened 2022-05-10
193 switch workflows to global

Yuri05:
Workflows which are the same in every repositories are now defined globally in 
https://github.com/Open-Systems-Pharmacology/Workflows and just referenced here.

Much better than copy-pasting every time :)


.github/workflows/MarkdownLinksCheck.yml
```diff
@@ -3,19 +3,6 @@ name: Check Markdown links
...
-    - uses: actions/checkout@v2
-      with:
-        ref: ${{ github.event.pull_request.head.sha }}
-    - uses: gaurav-nelson/github-action-markdown-link-check@v1
-      with:
-        use-quiet-mode: 'yes'
-        use-verbose-mode: 'yes'
-        check-modified-files-only: ${{ github.event_name == 'pull_request' }}
-        config-file: '.github/workflows/mlc_config.json'
+    uses: Open-Systems-Pharmacology/Workflows/.github/workflows/MarkdownLinksCheck.yml@main
```
Yuri05, line 8: That's all I need to do: refer to the global workflow (here using the global workflow in the **main** branch of the repo).
If I change the configuration of the global workflow - it will be automatically propagated to this repo, so no change required here

msevestre: Why are there so many files changed in this PR?

msevestre:
Not sure about a global wordlist....should we have a way to add some words at the repo level?

.github/workflows/wordlist.txt
```diff
@@ -1,1213 +1,3 @@
...
-GLP
-ables
-igures
-istings
-tlf
-reportingengine
-parameteridentification
+PKParameter
+PKSim
+userid
```
Yuri05, line 3: Spellchecker is now using 2 word lists:
* global word list under https://github.com/Open-Systems-Pharmacology/Workflows/blob/main/Data/wordlist_osp_global.txt 
* local word list (this one)

The idea is: words which are **always** valid, should be added to the global word list.
Words which sre valid only **in the context of the current repo**, should be added to the local word list.
E.g. **PKSim** is in general not valid (it should be **PK-Sim**), but in this repo it is used e.g. in "PKSim.CLI" etc. - thus adding it to the local word list.

Yuri05: >Not sure about a global wordlist....should we have a way to add some words at the repo level?

Yep, that's exactly what is described above. Global (= always valid) + local (valid for this repo only)

.github/workflows/MarkdownLinksCheck.yml
Yuri05: Just wrongly created PR into master instead into V11 first :)

PR #208, opened 2022-11-03
text and screenshots updated

### LINKED ISSUE #203: Mapping Outputs to Observed Data (v11.1): Replace Data with Observed Data in Text and scrrenshots
grafik

georgeDaskalakis:
closes #203


Yuri05:
it makes sense to update the screenshots only **after** the caption and the icon of observed data tab were changed in the software 😉

georgeDaskalakis:
> it makes sense to update the screenshots only **after** the caption and the icon of observed data tab were changed in the software 😉

hehe true, had an older version, didn't even notice.... correcting

georgeDaskalakis:
@Yuri05 corrected. The change is actually implemented in code, but only in Core, the dlls are not yet plugged in to PK-Sim or MoBi

Yuri05 (approved):

PR #214, opened 2022-12-21
Adding description for residual calculation

georgeDaskalakis:


part-3/pk-sim-simulations.md
```diff
@@ -187,6 +187,10 @@ For each mapping, the scaling can be defined as Lin or Log which determines the
 
 To reflect the quality or importance of the Observed Data item or to balance different magnitudes of values in case of Lin scaling, you can edit the weights of each mapping.
 
+### Handling of missing values for residuals
+
+If there is no simulated time value corresponding to an observed time value we use linear interpolation to calculate the missing simulated value. This happens in the calculation of the Residuals in the Simulation, opposed to the calculation of Residuals in Parameter Identification (where we insert an additional time point which has exactly the same value as the time value of the observed data).
```
PavelBal, line 192:
```suggestion
If there is no simulated time value corresponding to an observed time value, linear interpolation between simulated points is used to calculate the missing simulated value. This happens in the calculation of the Residuals in the Simulation, opposed to the calculation of Residuals in Parameter Identification (where every observed time point is added to simulation output points). This might result in slightly different total error values calculated in "Simulation" view and in Parameter Identification.
```

part-5/parameter-identification.md
```diff
@@ -158,6 +158,10 @@ If you select a mapping, the corresponding Observed Data is displayed as table a
 In MoBi®, you can freely define observers to match any kind of observed data. In order to enable the automatic mapping ensure that the meta data for Organ, Compartment and Molecule is the same for corresponding observed data and observers. Therefore, you can define the observers for specific Containers and Molecules and/or edit the meta data of the observed data.
 {% endhint %}
 
+### Handling of missing values for residuals
+
+If there is no simulated time value corresponding to an observed time value we insert an additional time point that has exactly the same value as the time value of the observed data. This happens in the calculation of the Parameter Identification Residuals, opposed to the calculation of Residuals in SImulation (in Simulation linear interpolation of the simulated values is used).
```
PavelBal, line 163:
```suggestion
For every observec data time point, the respective time value is added to the output interval of the simulation. This way, simulated value exists for each observed time point, and no interpolation of simulation results is required for the calculation of the total error.
```

Yuri05, line 163: typo: `SImulation `

Yuri05:
@georgeDaskalakis Can you fix (s. review comments)? After that we can merge I think

part-5/parameter-identification.md
```diff
@@ -160,7 +160,7 @@ In MoBi®, you can freely define observers to match any kind of observed data. I
 
 ### Handling of missing values for residuals
 
-If there is no simulated time value corresponding to an observed time value we insert an additional time point that has exactly the same value as the time value of the observed data. This happens in the calculation of the Parameter Identification Residuals, opposed to the calculation of Residuals in SImulation (in Simulation linear interpolation of the simulated values is used).
+For every observec data time point, the respective time value is added to the output interval of the simulation. This way, simulated value exists for each observed time point, and no interpolation of simulation results is required for the calculation of the total error.
```
Yuri05, line 163: typo: *observec*

part-5/parameter-identification.md
georgeDaskalakis: fixed

part-5/parameter-identification.md
georgeDaskalakis: fixed

georgeDaskalakis:
@Yuri05 cade review notes fixed. I still seem to be getting some errors when the tests run: https://github.com/Open-Systems-Pharmacology/docs/actions/runs/3891203461/jobs/6641153449

Yuri05:
@georgeDaskalakis All links are valid. 1 Link (below) is correctly reported as invalid, because it's indeed invalid on the level of source markdown files. However it's valid in the final rendered documentation (so apparently GitBook creates here a HTML anchro differently from GitHub). 
All other links are just false alarms - they are valid both in source MD as well as in the final rendered documentation. 
Probably some bugs in the link validator whatsoever.

So all good here

```
In file part-2/modules-philsophy-building-blocks.md
     bad reference (relative) at src:70:112-256:
       - text: "Exporting Project to Snapshot"
       - link: ../part-3/importing-exporting-project-data-models.md
       - anchor: exporting-project-to-snapshot-loading-project-from-snapshot
     
     ⛀  Anchor 'exporting-project-to-snapshot-loading-project-from-snapshot' is not present,
        did you mean exporting-project-to-snapshot--loading-project-from-snapshot (header II) at src:93:1-64?

```

Yuri05 (approved):

PR #222, opened 2023-03-24
221 editing description of LLOQ behaviour

### LINKED ISSUE #221: Improve description of LLOQ values
Here: https://docs.open-systems-pharmacology.org/shared-tools-and-example-workflows/import-edit-observed-data#lloq

If the LLOQ column is used - what are the units? What format should the entries be in - "1" or "<1" or "1 ng/ml" or?

What should be the corresponding entry in the "Measurement" column, if there is an entry in "LLOQ" column? Emtpy? "BLOQ", "<1", or what?

georgeDaskalakis:
closes #221


part-5/import-edit-observed-data.md
```diff
@@ -334,7 +334,13 @@ When setting the unit manually, the user needs to select the dimension first, up
...
+The LLOQ can either be specified from the column of the measurement or from a separate column. 
+
+In the first case (specified in the measurement column), the LLOQ values in the measurement column must be preceded by a "<", e.g. "<0.2", where 0.2 is the LLOQ value. 
+
+In the second case (specified in a separate column), the values in the column must not be preceded by "<" or anything else, neither should they have the unit written next to them in the cell: the LLOQ column should just contain arithmetic values. The unit in this case is always the same as the measurement unit. There can only be one single LLOQ value for every data set. In case there are several LLOQ values defined, the user is warned, and in case the user wants to proceed with the import, the highest of these LLOQs will be assumed for the whole data set.
```
PavelBal, line 341: Replace "arithmetic values" with "numerical values".

PavelBal, line 341: If LLOQ is defined in the LLOQ column, which values from the "Measurement" column are considered as below LLOQ? What happens when LLOQ column AND "<2" is entered in the Measurement column?

PavelBal (changes requested):

part-5/import-edit-observed-data.md
georgeDaskalakis: done

Yuri05 (approved):

PavelBal (approved):

PR #238, opened 2024-09-30
237 best practices section

rwmcintosh:


rwmcintosh:
@StephanSchaller FYI

rwmcintosh:
Still have a couple TODO to address. One link that was added here that doesn't work, but one that seems to be not working for some time.

Spell check and cross ref checks seem to be minimal, and are pointing to things that are actually not a problem

assets/images/part-7/model-development-1.png
Yuri05, line 1: @StephanSchaller The image has poor quality. Can we replace it with a one with better quality (ideally also SVG instead of PNG)

PR #245, opened 2025-01-14
Fixes #242 Examples are no longer delivered by the installer

### LINKED ISSUE #242: Examples are no longer delivered by the installer
And at least `Theophylline.pksim5` is not available in the model library

rwmcintosh:


rwmcintosh:
@Yuri05 any idea how we can see a preview on gitbook?

Yuri05 (approved):

PR #261, opened 2025-08-21
Part 3 populations

PavelBal:


Yuri05 (changes requested):
In PK-Sim, we use the term `gender`, so let's keep it in the docu as well and not change to `sex`

part-3/pk-sim-creating-populations.md
```diff
@@ -23,90 +24,77 @@ to the NHANES study](../assets/images/part-3/BW-vs-age.png)
...
+The population database contains physiological information starting from newborns preterms to newborns (age "0") to 80 year old individuals. When a pediatric or an elderly population is simulated, only the age dependence of anatomical properties (size of the organs, blood flow rates, body composition, etc.) is considered. However, important pharmacokinetic differences in children and elderly can be caused by age-related differences in, for example, hepatic metabolic activity. These effects are automatically accounted for if the ontogeny of the elimination process has been defined in the **Expression** tab for the underlying **Individual**. To manually account for variations in elimination, use the **User Defined Variability** tab.
```
Yuri05: ICRP: 100 years

PavelBal:
> In PK-Sim, we use the term `gender`, so let's keep it in the docu as well and not change to `sex`

I really think we should change it. We are repeatedly getting feedback that this is not up-to-date anymore.

Yuri05:
> I really think we should change it. We are repeatedly getting feedback that this is not up-to-date anymore.

If we change it one day in the PK-Sim UI, we should update the docu accordingly. But as long as we did not: let's keep the docu consistent with the software.

<details><summary>P.S.</summary>

If you check the number of hits on pubmed: _gender_ wins over _sex_ by far.
</details>

part-3/pk-sim-creating-populations.md
PavelBal: Done

PavelBal:
> In PK-Sim, we use the term `gender`, so let's keep it in the docu as well and not change to `sex`

Reverted.

part-3/pk-sim-creating-populations.md
```diff
@@ -210,7 +198,7 @@ The selected population will appear in the **Building Block Explorer** view.
 
 In addition, populations can be directly loaded from the template database within a simulation (see [PK-Sim® - Simulations](pk-sim-simulations.md).
 
-### Deleting Populations‌
+## Deleting Populations‌
 
 To delete a population from the project:
```
Yuri05, line 213: <details><summary>Please add the missing <I>Extract Individuals</I> menu description:</summary>

## Extracting individuals

Extract specific individuals from a population and add them as standalone building blocks to the project for follow-up analyses.
Individuals of interest can be identified within a population analysis (e.g. individuals associated with the percentiles displayed in a population Box-Whisker plot)

grafik
</details>

Yuri05: After this can be merged

PR #283, opened 2025-08-29
Part 4 ss bb

PavelBal:
I think different BBs should be described in separate documents, as it is done in PK-Sim.


part-4/model-building-components.md
```diff
@@ -6,6 +6,15 @@ After having made yourself familiar in the previous chapters with the building b
 Always watch for the helpful tool tips that appear when hovering for a few seconds with the mouse pointer over an input box or its description.
 {% endhint %}
 
+## Exporting and importing building blocks‌
```
PavelBal, line 9: This is for later rework. I think here should be a section that describes functionalities common for all BBs.

part-4/model-building-components.md
```diff
@@ -281,97 +290,6 @@ Models generated in **PK-Sim**® make extensive **use of tags**: For example, op
 
 Similarly, observers or events can be included or excluded from being created in different parts of the spatial structure. The molecule observer "Fraction excreted", for example, makes use of the tag "Urine", so this observer is only created in the urine container.
 
-## Spatial Structures‌
```
PavelBal, line 284: Moved to a separate document.

part-4/spatial-structures-bb.md
```diff
@@ -0,0 +1,108 @@
+# Spatial Structures‌
```
PavelBal, line 1: Aaaaah damn, but this way you don't see the changes @Yuri05 Well I reworked it completely, do you want to review it like this, or should I move it back to "model-building-components" first?

Yuri05: OK, I would suggest to merge this PR now AS IS and integrate the changes (if any) in later PRs

Yuri05 (approved):

PR #289, opened 2025-09-02
Part 4 - Observers BB

PavelBal:


part-4/model-building-components.md
```diff
@@ -498,9 +498,13 @@ If more than one molecule is transported by the very same transporter, you just
...
+An observer which can be displayed in a chart (see [Simulation Results](simulation-results.md)) is an output derived from one or several molecules or parameters by a defined formula. There are two classes of observers: **molecule observers** and **container observers**; creating and editing of both classes will be explained in this section. The main difference between those two classes is:
+
+- **Molecule observers** are calculated for instances of molecules in *physical containers* where these molecules are present.
+
+- **Container observers** can be calculated for *any container* (physical and logical) in the spatial structure, independent of whether a certain molecule is present in this container or not. However, even a container observer must be defined for at least one molecule, as it will be created as the property of this molecule.
```
PavelBal: @Yuri05 I still don't really understand the difference.

**Molecule observers** are clear - they are a property of a molecule instance in a physical container.

So, I first thought that **container observers** exist to create observers that should be a property of a logical container (e.g., total tissue concentration). But then, why is, for example, `Fraction of oral drug mass absorbed into mucosa segment` a container observer and not a molecule observer?

Yuri05: In principle, you are right, and container observers should be used when their parent is **not** a molecule amount.
However, any molecule observer can be defined as container observer with exactly the same combination of container criteria and molecule names - the resulting observer in a simulation would be the same in both cases.

So theoretically we could drop the `Molecule Observer` type and continue with the `(Container) Observer` only.
The [current implementation of container observer creation](https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/blob/12708fdefb4c3ee037e23d45d88afb49d85947bf/src/OSPSuite.Core/Domain/Services/ObserverBuilderTask.cs#L105-L126) would handle both use cases.

PavelBal: Honestly, I would be in favor of dropping the Molecule Observer type. I think this always has been a source of confusion for the users (including myself).

Yuri05: > Honestly, I would be in favor of dropping the Molecule Observer type. I think this always has been a source of confusion for the users (including myself).

No objections, let's create an issue :)

part-4/model-building-components.md
```diff
@@ -57,6 +57,13 @@ To save the current project select **Save Project (Ctrl+S)** in the File tab or
 
 You can then choose either <img src="../assets/icons/Save.svg" data-size="line">  **Save Project**, which saves the project under the name already given or <img src="../assets/icons/SaveAs.svg" data-size="line"> **Save As** to save the project with a new name in another file. The currently opened project will be saved with all information.
 
+## Keywords
```
PavelBal, line 68: To be described later.

part-4/events-bb.md
Yuri05, line 1: Empty file?

part-4/model-building-components.md
Yuri05: somehow, this **Keywords** section does not really fit in here for me.
It should be in the [section with the formulas below](#working-with-formulas) (where the references used in the formula are explained)

part-4/model-building-components.md
```diff
@@ -498,67 +505,84 @@ If more than one molecule is transported by the very same transporter, you just
...
 
 ![Select Relative Path window](../assets/images/part-4/EnteringRelativePath.png)
 
-1. On the right hand side of the edit window, the "Possible Referenced Objects" tree, "Vial1" is now highlighted. Open the "Vial1" tree by clicking on the + symbol to the left of "Vial1". You will see the parameter "Volume" below it, after "MoleculeProperties".
-2. Drag and drop exactly this "Volume" to the left, below the "M". The alias "Volume", the path "..|..|Volume", and the dimension "Volume" should appear. Compare the screen shot below with your monitor window.
-3. Finally, enter the formula "M/Volume" into the input box below the references (showing a red symbol <img src="../assets/icons/ErrorProvider.svg" data-size="line"> next to it before the formula is entered), and all should look like in this image.
+1. In the "Possible Referenced Objects" tree, navigate to `Organism|Vial1` and expand it. You will see the parameter `Volume` below it.
```
Yuri05: `Organism|Vial1` does not match the grafik above.
Also the enumeration counter is somehow corrupted

grafik

part-4/events-bb.md
PavelBal: Contents of the "Observers BB" will be moved to this file after approval. I do the changes in the "model-building-components" first so you can review the changes.

part-4/model-building-components.md
PavelBal: Changed to `BigVial|Vial1`. Added the screenshot to the list of screenshots to update.

part-4/model-building-components.md
PavelBal: Yes, just a placeholder for now :)

Yuri05 (approved):

PR #303, opened 2025-09-17
- Move descritpion how to save and load BBs into bb section

PavelBal:


part-4/model-building-components.md
```diff
@@ -1,28 +1,3 @@
-# Model Building and Model Components
```
PavelBal, line 1: This should also go, I think

PavelBal:
@Yuri05 please re-review (no pressure, just pinging).

Yuri05 (approved):

Yuri05:
Can be merged after resolving the conflicts

PR #316, opened 2025-11-04
Database extraction

PavelBal:


PavelBal:
> I am not sure, why the Postpartum population was selected for the example code.
Currently it is not part of the official PK-Sim release, so that applying the example code returns just empty tables.
E.g. the Pregnant population might be a better choice

@Rapha-L this is a valid comment. Can you re-write the documentation so that it uses a population that exists in PK-Sim? Otherwise user cannot follow your documentation because it would not work.

PR #329, opened 2026-04-23
Merge behavior molecules

PavelBal:
## Summary by CodeRabbit

* **Documentation**
  * Expanded and clarified merge rules for spatial structures, explicitly detailing how properties, parameters, container types, tags, and neighborhoods behave in Extend vs Overwrite modes.
  * Revised molecule semantics: added a dedicated "Molecules" section describing always-available molecules across modules and clearer rules for how molecule fields and parameter sets are handled in Extend and Overwrite modes.


part-4/modularization-concept.md
```diff
@@ -68,26 +68,46 @@ There are two types of combination behavior that can be defined for a module - *
 
 ### Spatial structure
 
+#### Merge behavior "Extend"
```
PavelBal, line 71: Here just swapped "Extend" and "Overwrite", so the default mode comes first.

Yuri05 (approved):

PR #336, opened 2026-06-02
Database extraction - update documentation to use pregnant population as it was suggested 

SoFiHo05:
I updated the documentation for time-dependent parameters to use the pregnant population as exampel, since this population is already integrated in PK-Sim/MoBi. This change has been suggested in an earlier PR

@PavelBal and @Rapha-L maybe you can check if it looks alright

## Documentation

* Updated database extraction workflow to focus on pregnancy instead of postpartum analysis
* Revised parameter definitions and population filtering criteria for pregnancy population
* Adjusted parameter detection logic with updated age intervals
* Modified import instructions and parameter guidance for pregnancy-specific scenarios
* Updated visualizations and references for pregnancy case studies


Yuri05:
@coderabbitai full review

Issue #335, opened 2026-04-24
V13: Merge Behavior Documentation Discrepancies


Yuri05:
This is the result of the comparison of the [current V13 documentation of merge/overwrite behavior](https://github.com/Open-Systems-Pharmacology/docs/blob/v13/part-4/modularization-concept.md#creating-simulations-from-modules-and-combination-rules) vs. the implementation in OSPSuite.Core (V13)

### Full report
https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/blob/copilot/research-merge-behavior-documentation/merge-behavior-findings.md

### Summary Table

| # | Claim / Topic | File & Lines | Status | Notes |
|---|---|---|---|---|
| 1 | `MergeBehavior` enum — two values, module-level, Overwrite default | `Module.cs` 18–34, 56 | ✅ Confirmed | Exactly two values; default = `Overwrite` |
| 2 | `analyzeBuilderMerges` dispatch, `tryExtendContainers`, 5 type-specific methods | `SimulationBuilder.cs` 96–135, 143–151, 164/172/224/261/266 | ✅ Confirmed | All present; reactions call `tryExtendContainers` twice (idempotent) |
| 3a | Spatial Structure — MoleculeProperties always extended (even in Overwrite) | `SpatialStructureMerger.cs` 81–93, 144 | ✅ Confirmed | Hard-coded special case for `MOLECULE_PROPERTIES` name |
| 3b | Spatial Structure — parameters replaced by name in Extend | `ContainerMergeTask.cs` 62 | ✅ Confirmed | `AddOrReplaceInContainer` for all leaf entities |
| 3c | Spatial Structure — container Mode overwritten | `ContainerMergeTask.cs` 67 | ✅ Confirmed | `targetContainer.Mode = containerToMerge.Mode` |
| 3d | Spatial Structure — tags behavior | `ContainerMergeTask.cs` 68 | ⚠️ Nuanced | Tags are **additive** (source tags appended, existing not removed) |
| 4a | Molecules — `QuantityType` NOT changed in Extend (doc claim) | `SimulationBuilder.cs` 274 | ❌ Discrepancy | Code **does** set `target.QuantityType = incoming.QuantityType` |
| 4b | Molecules — `IsFloating` behavior | `SimulationBuilder.cs` 272 | ❌ Discrepancy (if doc claims unchanged) | Code **does** set `target.IsFloating = incoming.IsFloating` |
[...]

PR #339, opened 2026-07-29
v12→v13 conversion guideline

PavelBal:
## What

Two changes to the MoBi documentation:

1. **Renamed** the section *"Reuse of Project Information from Previous Versions"* to **"Project conversion workflows"**. The page file (and therefore its URL) is unchanged; only the title and the `SUMMARY.md` entry change. The page now opens with a short overview pointing at the three conversion paths it covers (MoBi 2 / MoBi 3.1, ≤v11 → v12, v12 → v13).

2. **Added** a new sub-page under that section: [`part-4/converting-v12-projects-to-v13.md`](part-4/converting-v12-projects-to-v13.md) — *"Converting v12 projects to v13"*.

## Why the new page

`modularization-concept.md` on this branch already documents the **v13** merge rules correctly, but it does not say which of them **changed relative to v12**. That gap matters, because a model configuration authored in v12 can build a *different* simulation in v13 with no edit to the modules at all: for Molecules, Reactions, Passive Transports and Observers, v12 always did a full overwrite-by-name in *both* merge modes, so "Extend" and "Overwrite" were effectively the same. In v13 "Extend" genuinely extends, and "Overwrite" is the mode that reproduces the old behavior.

The new page therefore does not restate the rules as reference material — it links to `modularization-concept.md` as the authority and covers only:

- a per-building-block v12 vs. v13 comparison with a migration-impact rating,
- the failure modes this causes in practice, including the **silent loss of reactions** when educts/products are unioned instead of replaced, and the **build failure when combining two PK-Sim modules** in v13 (`'<Molecule>|Is small molecule' that cannot be found`) with the fix (set the PK-Sim modules to "Extend"),
- a step-by-step migration workflow based on building the simulation in both versions and comparing the PKML exports semantically,
- an observed **discrepancy between the documented and actual v13 event merge behavior**: for an equally-named application administering different molecules, v13 "Extend" keeps the *first* module's molecule rather than extending to both as documented.

The behavioral claims come from controlled two-module tests built in both v12 and v13 and compared via PKML export; notably, the v13 "Extend" build of two PK-Sim modules reproduced the v12 "Overwrite" result exactly.

A second, unrelated v13 migration blocker is documented at the end: the new oral absorption model rejects strongly negative Lipophilicity values (MoBi #2445).

## Review notes

- The event-behavior section states that v13 does not match the documented rules. If this is a known/intended change rather than a bug, that section should be reworded — please flag it.
- The v13 "Overwrite" case for equally-named events administering different molecules was not tested in isolation and is marked as such in the table.

## Summary by CodeRabbit

* **Documentation**
  * Added guidance for converting v12 projects to v13, including updated module merge behavior and migration steps.
  * Documented changes affecting molecules, reactions, transports, observers, spatial structures, and events.
  * Added notes on PK-Sim module conversion, low or negative lipophilicity, and potential simulation differences.
  * Updated modularization guidance to clarify molecule type handling in Extend and Overwrite modes.
  * Organized project conversion workflows and expanded the documentation table of contents.

### DIFF
--- a/part-4/converting-v12-projects-to-v13.md
+++ b/part-4/converting-v12-projects-to-v13.md
@@ -137,7 +137,7 @@ In a controlled test with two large-molecule PK-Sim modules, the **v13 "Extend" 
 The container/parameter/tag/neighborhood rules are essentially as in v12, with two clarifications added in v13:
 
 - **`MoleculeProperties` are now always extended** in *both* "Extend" and "Overwrite" modes: new molecule properties from the later module are added, and a property present in both modules takes the later module's value/formula.
-- For neighborhoods under **"Extend"**, v13 clarifies that **neighbors are replaced** by the later module; v12 only stated that neighborhoods are extended. (Under "Overwrite", v12 already specified that neighbors are overwritten.) Both modes now also carry a warning: if the later module defines a neighborhood with *invalid* neighbors, the earlier module's neighborhood is kept unchanged — neighborhoods cannot be removed from a model.
+- For neighborhoods under **"Extend"**, v13 clarifies that **neighbors are replaced** by the later module; v12 only stated that neighborhoods are extended. (Under "Overwrite", v12 already specified that neighbors are overwritten.) Both modes now also carry a warning: if the later module defines a neighborhood with *invalid* neighbors, the earlier module's neighborhood is kept unchanged — neighborhoods cannot be removed from a model. This behavior is tracked in the issue https://github.com/Open-Systems-Pharmacology/MoBi/issues/2367 and will be changed before the official v13 release.
 
 **Migration impact — low to medium**, depending on whether molecule-property values/formulas were relied upon to be replaced rather than merged.
 
@@ -154,7 +154,9 @@ Observed in a controlled two-module test (module A administers molecule X, modul
 
 Under v13 "Extend" the administered molecule is therefore *not* combined, and it follows the **opposite** precedence to parameters, where the later module wins. The current rules are documented in [Events](modularization-concept.md#events).
 
-**Migration impact — medium.** A v12 Extension module that redefined the administered molecule of an existing application under "Extend" now has **no effect at all** — the base module's molecule is kept, silently. Note also that event combination is expected to change again in a future release.
+This first-module precedence is considered a bug — the administered molecule should be taken from the later module, as every other overwritten property is. It is tracked in https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/issues/2917.
+
+**Migration impact — medium.** A v12 Extension module that redefined the administered molecule of an existing application under "Extend" now has **no effect at all** — the base module's molecule is kept, silently. Note also that event combination is expected to change again in a future release, both through the fix for the precedence bug above and through the planned [application definition rework](https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/issues/2128).
 
 {% hint style="warning" %}
 **Avoid defining events/applications with the same name in more than one module.** Where this is unavoidable, explicitly verify **which molecule the application administers** after building the simulation.
@@ -167,13 +169,37 @@ Under v13 "Extend" the administered molecule is therefore *not* combined, and it
 
 ## Migration steps
 
-The goal is to confirm that a v12 model configuration produces the *same* simulation in v13, and to correct it where the new merge rules change the result.
+The goal is to arrive at a v13 model configuration whose every difference from the v12 original is **understood and intended**. Two distinct sources of difference have to be separated: the v13 **PK-Sim PBPK model structure**, whose changes you generally *want* to adopt, and the new **merge rules**, whose effects are usually unintended and must be corrected. The steps below deal with them in that order.
 
-1. **Inventory the model configuration.** List every module in the configuration, its position in the hierarchy (order of selection), and its merge mode ("Extend" / "Overwrite").
+1. **Inventory the model configuration.** List every module in the configuration, its position in the hierarchy (order of selection), and its merge mode ("Extend" / "Overwrite"). Note which of them are **PK-Sim modules** and which are MoBi **Extension modules**, and record the PK-Sim version each PK-Sim module was created with.
 
-2. **Flag the affected building blocks.** For each Extension module, identify whether it redefines any **molecule, reaction, passive transport, or observer** that also exists in a module higher in the hierarchy. These are the entities whose merge result may have changed. **Also flag any event/application defined with the same name in more than one module** — especially administering *different* molecules; verify the administered molecule in the built simulation (see [Events](#events)), and prefer to avoid equally-named events across modules altogether.
+2. **Re-create all PK-Sim modules in PK-Sim v13.** Do this *before* any merge-mode work — everything downstream is built on these modules, so changing them later means redoing the comparisons.
 
-3. **Establish the ground truth by comparison.** For each affected simulation, build the simulation in **both** v12 and v13 and export the result to `*.pkml`. Compare the two PKML files. This reveals exactly which entities merged differently — it is far more reliable than reasoning from the rules alone.
+   A PK-Sim module created in v12 is a v12 artefact: it carries the v12 PBPK model structure and the v12 physiological and molecular database. Simply carrying it into a v13 project keeps all of that, and none of the v13 improvements are applied — the same reasoning as for [PK-Sim projects](../part-3/importing-exporting-project-data-models.md#exporting-project-to-snapshot--loading-project-from-snapshot). Each PK-Sim module must therefore be re-created in PK-Sim v13, re-imported into the MoBi project in place of the old one (**preserving each simulation's module order**), and **every** simulation that uses it rebuilt.
+
+   - If the original **PK-Sim project or snapshot still exists**, export it to a snapshot and load that snapshot in PK-Sim v13. All v12 → v13 conversions are applied on load. This is the recommended path and requires no manual re-entry of inputs.
+   - If **neither exists**, the module has to be rebuilt by hand in PK-Sim v13. Origin data, compound parameters, expression profiles, applications and calculation methods can be read off the v12 module, but which PK-Sim *input choice* produced a given value generally cannot — those inputs must come from the original project or from your notes.
+
+   {% hint style="info" %}
+   Starting with v13, a PK-Sim module **embeds its PK-Sim snapshot**, so a v13 module carries everything needed to re-create itself. v12 modules do not, which is why this step is manual today. **Export a PK-Sim snapshot and store it alongside the model** as part of this migration — it is what makes the next migration straightforward.
+   {% endhint %}
+
+   After re-importing, check the **merge behavior** of the new PK-Sim modules before rebuilding — see the note on [combining PK-Sim modules](#passive-transports) above.
+
+3. **Compare simulations built from the PK-Sim modules only.** Before looking at the full model configuration, build a simulation containing **only** the PK-Sim modules — the v12 module in v12 and the re-created v13 module in v13 — and compare the two exports.
+
+   Every difference found here originates from **changes in the PK-Sim PBPK model structure**, not from the modularization merge rules. Establishing that list separately is what allows you to attribute the *remaining* differences in the full configuration to merge behavior instead of guessing which of the two causes is responsible.
+
+   Differences commonly seen at this step, which come from the new v13 oral absorption model and the building-block renaming rather than from anything in your model:
+
+   - additional per-compound **bile-salt-micelle** parameters (critical micellar concentration, partition coefficient bile-salt-micelle/water, …)
+   - additional per-`Lumen`-compartment parameters (aqueous solubility, fluid velocity and viscosity, micellar diffusion, …)
+   - **rewritten intestinal `Solubility` formulas** — these are *not* cosmetic and are numerically relevant for any oral administration
+   - building-block renames `Reaction` → `Reactions` and `Observer` → `Observers`
+
+4. **Flag the affected building blocks.** For each Extension module, identify whether it redefines any **molecule, reaction, passive transport, or observer** that also exists in a module higher in the hierarchy. These are the entities whose merge result may have changed. **Also flag any event/application defined with the same name in more than one module** — especially administering *different* molecules; verify the administered molecule in the built simulation (see [Events](#events)), and prefer to avoid equally-named events across modules altogether.
+
+5. **Establish the ground truth by comparison.** Now for the **full** model configuration: build each affected simulation in **both** v12 and v13 and export the result to `*.pkml`. Compare the two PKML files, discounting the PK-Sim-only differences already catalogued in step 3. What remains reveals exactly which entities merged differently — it is far more reliable than reasoning from the rules alone.
 
    {% hint style="info" %}
    A raw text diff of two PKML exports is not useful — element IDs and internal references are regenerated on every export. Compare **semantically** instead: compare entities by name and by container path, and compare counts per entity type. In particular, watch for **reactions/transports/observers whose *definitions* are identical but whose number of *instantiations* in the model tree differs** — that is the signature of the merge-rule change described above.
@@ -181,13 +207,13 @@ The goal is to confirm that a v12 model configuration produces the *same* simula
    Exported simulations may contain `SimulationEntitySource` provenance records (each entity traced to its origin `moduleName`, `buildingBlockType`, `sourcePath`). These do **not** affect model behavior. Their presence and count depend on the MoBi/export build rather than on model semantics — the same OSP version can export with or without them — so **do not treat them as a v12↔v13 difference; ignore them when comparing.**
    {% endhint %}
 
-4. **Decide per module: Extend or Overwrite.**
+6. **Decide per module: Extend or Overwrite.**
    - If the module was meant to **fully replace** a molecule/reaction/transport/observer (the common v12 assumption), set its merge mode to **"Overwrite"**.
    - If it was meant to **add to** an existing building block, keep **"Extend"** and verify the merged result — remove any now-redundant duplicated content, and remember that reaction educts/products cannot be removed (use stoichiometry `0`).
 
-5. **Re-verify.** Rebuild in v13 after each change and compare against the v12 PKML again, until the simulation matches the intended v12 result (or until any intentional differences are understood and documented).
+7. **Re-verify.** Rebuild in v13 after each change and compare against the v12 PKML again, until the simulation matches the intended v12 result (or until any intentional differences are understood and documented).
 
-6. **Simplify.** Because "Extend" now merges rather than replaces, Extension modules that previously carried *complete* redefinitions (to satisfy the old full-overwrite rule) can often be reduced to only the differences — the recommended best practice. Do this only after the configuration is verified.
+8. **Simplify.** Because "Extend" now merges rather than replaces, Extension modules that previously carried *complete* redefinitions (to satisfy the old full-overwrite rule) can often be reduced to only the differences — the recommended best practice. Do this only after the configuration is verified.
 
 {% hint style="warning" %}
 When combining PBPK models with different structures, small-molecule models must still be selected **before** large-molecule models, or simulation creation fails due to missing parameters. This constraint is unchanged from v12.
--- a/part-4/converting-v12-projects-to-v13.md
+++ b/part-4/converting-v12-projects-to-v13.md
@@ -34,7 +34,7 @@ The table below summarizes the behavioral changes. Only **Parameter Values** and
 | **Passive transports** | "Extend" was **identical to** "Overwrite". | "Extend" and "Overwrite" are now distinct. |
 | **Observers** | "Extend" was **identical to** "Overwrite". | "Extend" and "Overwrite" are now distinct. |
 | **Spatial structure** | No explicit `MoleculeProperties` rule. | `MoleculeProperties` are now **always extended** in both modes; neighborhood "neighbors are replaced" behavior clarified. |
-| **Events** | Documented rules unchanged, but for an equally-named event administering *different* molecules: Overwrite → last module's molecule; Extend → both (malformed). | Same *documented* rules, but the actual "Extend" behavior keeps the **first** module's molecule (see below). |
+| **Events** | For an equally-named event/application administering *different* molecules, "Extend" extended the administered molecule to **both** (a malformed event). | "Extend" keeps the **first** module's molecule; a later module's redefinition is silently dropped (see below). |
 
 ### Molecules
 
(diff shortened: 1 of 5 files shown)


part-4/converting-v12-projects-to-v13.md
```diff
@@ -0,0 +1,202 @@
...
+  - **Parameters** list is **extended**.
+  - **Source** and **target** lists are **extended**; their operators are overwritten.
+  - **Include/Exclude** molecule lists are extended; the **"All" checkbox** behavior is overwritten (an exclusion in the later module takes precedence).
+- **Merge behavior "Overwrite"**
+  - The passive transport is **completely overwritten by name**.
+
+**Migration impact — medium.** Under v12, an "Extend" module dropped source/target/parameter entries that were not present in it. Under v13 "Extend", those entries are now **retained** from the earlier module. If the intent was replacement (including removal of entries), switch to **"Overwrite"**.
+
+{% hint style="danger" %}
+**Combining multiple PK-Sim modules can now fail to build.** PK-Sim modules are imported with the default merge behavior **"Overwrite"**. In v12, combining two PK-Sim modules still worked — even under "Overwrite", parts of passive transports and observers (notably the molecule **include/exclude lists**) were extended rather than replaced. In v13, "Overwrite" replaces those lists too, so a passive transport defined for one module's molecules loses the other module's molecules.
```
PavelBal: This behavior is changed in v13:

https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/3635 So only for v12, starting from v13, the default behavior is "Extend"

part-4/converting-v12-projects-to-v13.md
```diff
@@ -0,0 +1,202 @@
...
+
+This breaks the combination of a **large-molecule model with a small-molecule model**: the large-molecule transports (e.g. FcRn-mediated `NetMassTransfer_*`) end up referencing molecule-specific entities that are no longer created, and **simulation creation fails** with errors such as:
```
PavelBal: This is actually not a problem of the combination of large- and small molecules models. This error appears when combining two LM models, because the lists of "exclude" molecules are not combined but overwritten, when merge behavior is "overwrite".

part-4/converting-v12-projects-to-v13.md
```diff
@@ -0,0 +1,202 @@
...
+
+```
+Transport 'NetMassTransfer_InterstitialToEndosomal' references an entity with path
+'<Molecule>|Is small molecule' that cannot be found
+```
+
+**Fix:** set the PK-Sim modules' merge behavior to **"Extend"** (via MoBi, or `module$mergeBehavior <- "Extend"` in the `ospsuite` R package). With PK-Sim modules extended rather than overwritten, the molecule lists combine and the simulation builds. Note the general rule still applies: select small-molecule models **before** large-molecule models.
+
+Strictly, only the **later** module's mode matters: merge behavior governs how a module is combined into what precedes it, so the first module has nothing to overwrite and its mode is immaterial. Setting every PK-Sim module to "Extend" is the simpler rule and does no harm.
```
PavelBal: No. If more than 2 modules are combined, then the modes of all modules matter

part-4/converting-v12-projects-to-v13.md
```diff
@@ -0,0 +1,202 @@
...
+- For neighborhoods, v13 clarifies that **neighbors are replaced** by the later module, and adds a warning: if the later module defines a neighborhood with *invalid* neighbors, the earlier module's neighborhood is kept unchanged (neighborhoods cannot be removed from a model).
+
+**Migration impact — low to medium**, depending on whether molecule-property values/formulas were relied upon to be replaced rather than merged.
+
+### Events
+
+The **documented** event merge rules are unchanged from v12: events combine only when generated in the same container (by container criteria); under **Overwrite** the administered molecule is overwritten (last module wins), under **Extend** it is extended (both molecules, which results in a malformed event).
+
+However, the **actual v13 behavior does not match the documented rules** when two modules define an **equally-named event/application for *different* molecules**. Observed in a controlled two-module test (module A administers molecule X, module B — later in the hierarchy — administers molecule Y under the same application name):
```
PavelBal: Separate PR to correct the documentation

part-4/converting-v12-projects-to-v13.md
PavelBal: Fixed in 243d7eb. The paragraph now says "Overwrite" was the **v12** default and that v13 changes the PK-Sim module default to "Extend", linking PK-Sim#3635, and it frames the failure as affecting **model configurations carried over from v12** whose modules still carry the old default (rather than new v13 imports).

part-4/converting-v12-projects-to-v13.md
PavelBal: Fixed in 243d7eb. Reworded to "This surfaces when combining **two large-molecule models**" and attributes the cause to the molecule include/exclude lists being replaced instead of combined under "Overwrite". I also corrected the error message to the real one from PK-Sim#3635 (`<Molecule>-FcRn_Complex|Is small molecule`) and removed the now-misleading "select small-molecule before large-molecule" note from that hint — that constraint is unrelated to this failure and is still stated separately in the migration steps. The verification note further down now says the controlled test used two large-molecule PK-Sim modules, matching the issue.

part-4/converting-v12-projects-to-v13.md
PavelBal: You are right — fixed in 243d7eb. The sentence now reads: "Only the **first** module in the hierarchy has nothing to merge into, so its mode has no effect. The mode of every module selected after it matters — with more than two modules, check them all rather than only the last."

part-4/converting-v12-projects-to-v13.md
PavelBal: Agreed, and it turns out that separate PR already exists and is merged: docs#337 (commit 41fcdc5) corrected the Events → "Extend" section of `modularization-concept.md` on this branch. I had written this page against a stale local `v13` from before that commit, so the "actual behavior does not match the documented rules" framing was wrong.

Rewritten in 243d7eb: the Events section now presents only the v12 → v13 behavior delta (Extend: both molecules → first module’s molecule only, later one silently dropped), notes that Overwrite is unchanged, and links to modularization-concept.md#events as the authority instead of claiming the docs are wrong. Dropped the MoBi issue-tracker link, the "appears to be a bug" wording, and the "(not tested in isolation)" cell — per docs#337 the Overwrite case behaves as documented in v13, so the table now states that. Also updated the corresponding row in the summary table at the top of the page and added a migration-impact note: a v12 module that redefined an application’s administered molecule under "Extend" now has no effect at all.

PavelBal:
## Full claim audit against live `v13`

Since I admitted the page was drafted against a stale `v13`, I re-checked **every** claim. Method: `git diff origin/v12 origin/v13 -- part-4/modularization-concept.md`, which is the authoritative record of what the documented behavior change actually is, plus the upstream issues/PRs and the R sources.

**Scope of the staleness was limited to Events.** `git diff 5e9b4ea origin/v13 -- part-4/` shows the *only* change between the copy I drafted against and live `v13` is the administered-molecule bullet + warning from #337 — already fixed in 243d7eb. My branch is level with `origin/v13` (0 commits behind).

**Confirmed against the v12↔v13 diff:**

| Claim | Verdict |
|---|---|
| Molecules: v12 "always overwritten by name", redefinition needed all parameters | ✅ verbatim in v12 |
| Reactions: v12 "always overwritten by name" in both modes | ✅ verbatim in v12 |
| Passive transports: v12 "Extend" identical to "Overwrite"; parameters and source/target lists overwritten so entries could be removed | ✅ verbatim in v12 |
| Observers: v12 "Extend" identical to "Overwrite" | ✅ verbatim in v12 |
| Spatial structure: no explicit `MoleculeProperties` rule in v12 | ✅ the bullets are added in v13, in both modes |
| v12 "Overwrite" still extended the include/exclude molecule lists of PTs and observers (the premise of the build failure) | ✅ verbatim in v12, for both PT and observers |
| Events: v12 "Extend" extended the administered molecule to both (malformed) | ✅ verbatim in v12 |
| Events "Overwrite" unchanged between v12 and v13 | ✅ that bullet is untouched by the diff |
| Parameter Values and Initial Conditions resolution orders unchanged | ✅ neither section appears in the diff |
| "Small-molecule before large-molecule" constraint unchanged from v12 | ✅ best-practices bullet untouched by the diff |
| All v13 "Extend"/"Overwrite" bullet lists for molecules, reactions, PTs, observers | ✅ match live `modularization-concept.md` |
| `module$mergeBehavior <- "Extend"` | ✅ `MoBiModule$mergeBehavior` is a settable field validated against the `MergeBehavior` enum keys (`R/MoBiModule.R:26`) |
| MoBi#2445 | ✅ open, titled "v13: New oral model does not allow low Lipophilicity values" |
| All 7 cross-page anchors + the one self-anchor | ✅ all resolve against live headings |

**Three imprecisions found and fixed in fc91ce0:**

1. I wrote that molecule parameters under "Extend" are overwritten *"per absolute path"*. The docs say parameters are overwritten with the example `MolA|Param` and do not use that wording — reworded to describe the merge without inventing a matching rule.
2. The neighborhood *"neighbors are replaced"* clarification is specifically on the **"Extend"** side; v12 already stated that neighbors are overwritten under "Overwrite". The page now says so instead of implying the whole rule is new.
3. On PK-Sim#3635: the fix (PK-Sim#3636, `SimulationConfigurationTask.CreateFor`) sets `MergeBehavior.Extend` when a module is **created** from a PK-Sim simulation — it does not retroactively change modules already stored in a project. The page now states this explicitly, which is what makes the "carried over from v12" framing correct. Worth noting there is no v13 release yet (latest published is v12.3.173), so this ships with v13.

**One judgement call left for you:** the "malformed event uses a single molecule's molecular weight" detail and the PKML-comparison observations come from your controlled tests, not from anything in the repo, so I could not verify them here and have left them as written.

PavelBal:
Two commits added, both about the **molecule type** merge behavior.

The v13 documentation stated that under merge behavior "Extend" the molecule type is *not* changed. That is incorrect — `SimulationBuilder.mergeMolecules` assigns `QuantityType` unconditionally, so the type is taken from the later module in **both** modes. It only *looked* different because the molecule icon was not copied along with the type; that was fixed in Open-Systems-Pharmacology/OSPSuite.Core#2853.

- `part-4/modularization-concept.md` — corrects the "Extend" bullet. This one-line change is also submitted standalone in #340 (base `v13`), since it is a fix to already-written v13 documentation and independent of this PR. If #340 is merged first the change is identical on both sides and will merge cleanly; it is included here so this branch is internally consistent.
- `part-4/converting-v12-projects-to-v13.md` — removes the molecule type from the v12→v13 migration hazards. In v12 molecules were always fully overwritten by name, so the type was taken from the later module there too. It is therefore not a v12→v13 difference; what actually changes for molecules is the treatment of **parameters** and **active transports**.

Refs #335 (item 4a)

Yuri05:
@coderabbitai full review

Yuri05:
@PavelBal Please check the CodeRabbit comments above and also the remarks by Perplexity below

<details><summary>Perplexity  PR review</summary>
<br>

The PR is conceptually strong and mostly accurate, but there are a few concrete issues you should address before merging: some technical claims were later corrected elsewhere in the docs, the migration checklist is slightly incomplete, and a couple of minor tooling-related nits (spellcheck, markdownlint) are still open. [[github](https://github.com/Open-Systems-Pharmacology/docs/pull/339)](https://github.com/Open-Systems-Pharmacology/docs/pull/339)

Below I’ll focus on what’s *still* problematic or potentially confusing, given the PR’s own audit and follow‑up comments.

***

## 1. Molecule type behavior no longer a v12→v13 difference

The latest comments and commits in the PR note that the molecule type (QuantityType) is overwritten by the later module in **both** merge modes, via `SimulationBuilder.mergeMolecules` assigning `QuantityType` unconditionally. [[github](https://github.com/Open-Systems-Pharmacology/docs/pull/339)](https://github.com/Open-Systems-Pharmacology/docs/pull/339)

This has two implications:

- The earlier claim that under v13 “Extend” the molecule type is **not** changed is incorrect and has already been fixed in `part-4/modularization-concept.md` and clarified in a separate PR (#340). [[github](https://github.com/Open-Systems-Pharmacology/docs/pull/339)](https://github.com/Open-Systems-Pharmacology/docs/pull/339)
- Any mention of molecule type as a **migration hazard specific to v12→v13** is now wrong, because v12 also always overwrote molecules by name, including the type. [[github](https://github.com/Open-Systems-Pharmacology/docs/pull/339)](https://github.com/Open-Systems-Pharmacology/docs/pull/339)

You already partially addressed this by:

- Removing molecule type from the v12→v13 migration hazards in `part-4/converting-v12-projects-to-v13.md`. [[github](https://github.com/Open-Systems-Pharmacology/docs/pull/339)](https://github.com/Open-Systems-Pharmacology/docs/pull/339)
- Keeping that page focused on parameters and active transports (which *do* change). [[github](https://github.com/Open-Systems-Pharmacology/docs/pull/339)](https://github.com/Open-Systems-Pharmacology/docs/pull/339)

Remaining checks to perform in this PR:

- Re-scan the v12→v13 conversion page for **any residual references** that imply molecule type behaves differently between v12 and v13 (including summary tables, hints, and migration-impact paragraphs). They should all be removed or reworded so that only parameters and active transports are called out as changed.  
- Verify that the summary table row for **Molecules** now says something like: “Parameters and active transports behavior changed; type overwritten by later module in both modes (unchanged vs v12).” If it still suggests a type difference, that’s a bug.

***

## 2. Migration checklist: spatial structures and events still under-specified

CodeRabbit correctly flagged that the migration workflow steps (4–6) currently focus only on molecules, reactions, passive transports, and observers, even though the same document identifies spatial-structure behavior changes (MoleculeProperties and neighborhoods) and important event redefinition deltas. [[github](https://github.com/Open-Systems-Pharmacology/docs/pull/339)](https://github.com/Open-Systems-Pharmacology/docs/pull/339)

Issues:

- Step 4 (“Flag the affected building blocks”) does not explicitly mention **spatial structures**, despite the earlier section explaining that MoleculeProperties are always extended in v13 and that neighborhood handling under “Extend” has clarifications. [[github](https://github.com/Open-Systems-Pharmacology/docs/pull/339)](https://github.com/Open-Systems-Pharmacology/docs/pull/339)
- Events are mentioned, but the guidance is mostly “prefer to avoid equally-named events”; it lacks a clear, actionable “verify and decide Extend vs Overwrite” sub-step for applications that differ only in administered molecule. [[github](https://github.com/Open-Systems-Pharmacology/docs/pull/339)](https://github.com/Open-Systems-Pharmacology/docs/pull/339)

Concrete improvements you should make in `part-4/converting-v12-projects-to-v13.md`:

1. Expand step 4:
   - Include **spatial structures** explicitly (containers, MoleculeProperties, neighborhoods) in the list of building blocks to flag when they are redefined in later modules.  
   - Keep the existing emphasis that MoleculeProperties are now always extended (both modes) and neighborhoods under “Extend” may be replaced, so users know why these matter.

2. Add an explicit events/applications sub-step:
   - In either step 4 or 6, add a bullet such as:  
     - “For each application/event name that occurs in more than one module, build v12 and v13 simulations and verify which molecule is actually administered in each. Decide whether the later module should use **Extend** (combined or first-molecule behavior in v13) or **Overwrite** (last module wins), and document that choice.”  
   - Make the wording consistent with the corrected v13 event rules in `modularization-concept.md#events`, which the PR itself references. [[github](https://github.com/Open-Systems-Pharmacology/docs/pull/339)](https://github.com/Open-Systems-Pharmacology/docs/pull/339)

3. Extend step 6 (“Decide per module: Extend or Overwrite”) to:
   - Cover **spatial structures** explicitly, indicating when you want MoleculeProperties merged vs replaced, and reminding that neighborhoods cannot be removed and invalid neighbors lead to keeping the earlier definition. [[github](https://github.com/Open-Systems-Pharmacology/docs/pull/339)](https://github.com/Open-Systems-Pharmacology/docs/pull/339)
   - Cover **events** explicitly (e.g., “If the module was meant to change the administered molecule for an existing application name, use Overwrite; if it was meant to add an additional application that should not conflict by name, prefer distinct names.”).

This closes the gap CodeRabbit marked as a functional correctness issue. [[github](https://github.com/Open-Systems-Pharmacology/docs/pull/339)](https://github.com/Open-Systems-Pharmacology/docs/pull/339)

***

## 3. Spellcheck failures and terminology alignment

Your CI spellcheck currently rejects several terms used in the new page: `modularization`, `micellar`, `instantiations`, `redefinitions`, and `unioned`. [[github](https://github.com/Open-Systems-Pharmacology/docs/pull/339)](https://github.com/Open-Systems-Pharmacology/docs/pull/339)

The bot suggests:

- Adding **modularization**, **micellar**, **instantiations**, and **redefinitions** to the project’s approved terminology dictionary, as they are standard in OSP and PBPK. [[github](https://github.com/Open-Systems-Pharmacology/docs/pull/339)](https://github.com/Open-Systems-Pharmacology/docs/pull/339)
- Replacing **unioned** with **combined** where the meaning is unchanged. [[github](https://github.com/Open-Systems-Pharmacology/docs/pull/339)](https://github.com/Open-Systems-Pharmacology/docs/pull/339)

Issues:

- Until those dictionary updates are made, the CI will keep failing on the new document and any referenced sections, blocking merge.  
- “Unioned” is potentially confusing jargon; “combined” or “union of” is clearer, and matches the underlying concept of unioning educt/product sets across definitions.

Concrete actions:

- Update the project dictionary to include these domain terms, particularly **modularization**, which is central to this page. [[github](https://github.com/Open-Systems-Pharmacology/docs/pull/339)](https://github.com/Open-Systems-Pharmacology/docs/pull/339)
- In the “silent loss of reactions” example, replace “the products are unioned” with “the products are **combined** into a union of both sets”, which preserves meaning and satisfies the spellcheck recommendation. [[github](https://github.com/Open-Systems-Pharmacology/docs/pull/339)](https://github.com/Open-Systems-Pharmacology/docs/pull/339)

Double-check that similar wording appears only in this page and that you’ve updated all occurrences (CodeRabbit mentions lines around 79–81 and 195–205). [[github](https://github.com/Open-Systems-Pharmacology/docs/pull/339)](https://github.com/Open-Systems-Pharmacology/docs/pull/339)

***

## 4. Markdownlint: fenced code block without language

Markdownlint (MD040) is complaining about the transport error message code fence. [[github](https://github.com/Open-Systems-Pharmacology/docs/pull/339)](https://github.com/Open-Systems-Pharmacology/docs/pull/339)

Current block:

``` 
Transport 'NetMassTransfer_InterstitialToEndosomal' references an entity with path
'<Molecule>-FcRn_Complex|Is small molecule' that cannot be found
```

Issue:

- The block lacks a language identifier, which triggers MD040. [[github](https://github.com/Open-Systems-Pharmacology/docs/pull/339)](https://github.com/Open-Systems-Pharmacology/docs/pull/339)

Fix:

- Change the fence to:

```text
Transport 'NetMassTransfer_InterstitialToEndosomal' references an entity with path
'<Molecule>-FcRn_Complex|Is small molecule' that cannot be found
```

This is mechanically simple and clears the linter warning.

***

## 5. Checklist semantics and multi-module hierarchy nuance

You already corrected one earlier oversimplification: the claim that only the **later** module’s mode matters was wrong when more than two modules are combined. [[github](https://github.com/Open-Systems-Pharmacology/docs/pull/339)](https://github.com/Open-Systems-Pharmacology/docs/pull/339)

Current clarification:

- “Only the first module in the hierarchy has nothing to merge into, so its mode has no effect. The mode of every module selected after it matters — with more than two modules, check them all rather than only the last.” [[github](https://github.com/Open-Systems-Pharmacology/docs/pull/339)](https://github.com/Open-Systems-Pharmacology/docs/pull/339)

Remaining considerations:

- Verify that this nuance is consistently reflected in the migration checklist:  
  - Step 4 (flagging affected building blocks) should encourage scanning **all modules after the first**, not just the last one.  
  - Step 6 (“Decide per module: Extend or Overwrite”) should be clearly “per module after the first”, not implicitly “pick one global mode”.  

If any wording still suggests that setting all PK-Sim modules to Extend is always harmless or that only the last module’s mode matters, it needs tightening, especially for configurations with more than two modules. [[github](https://github.com/Open-Systems-Pharmacology/docs/pull/339)](https://github.com/Open-Systems-Pharmacology/docs/pull/339)

***

## 6. Overall technical soundness

From the internal audit section and the v12↔v13 diff review included in the PR, most of the per-building-block behavioral claims now match live v12 and v13 docs: [[github](https://github.com/Open-Systems-Pharmacology/docs/pull/339)](https://github.com/Open-Systems-Pharmacology/docs/pull/339)

- v12 “always overwritten by name” for molecules and reactions is correct. [[github](https://github.com/Open-Systems-Pharmacology/docs/pull/339)](https://github.com/Open-Systems-Pharmacology/docs/pull/339)
- v12 “Extend identical to Overwrite” for passive transports and observers is correct. [[github](https://github.com/Open-Systems-Pharmacology/docs/pull/339)](https://github.com/Open-Systems-Pharmacology/docs/pull/339)
- Spatial-structure clarifications about MoleculeProperties being extended in both modes and neighborhood replacement under Extend are correctly drawn from the v13 diff. [[github](https://github.com/Open-Systems-Pharmacology/docs/pull/339)](https://github.com/Open-Systems-Pharmacology/docs/pull/339)
- The PK-Sim module default merge behavior change (v12 Overwrite → v13 Extend at module creation, not retroactive) is correctly characterized. [[github](https://github.com/Open-Systems-Pharmacology/docs/pull/339)](https://github.com/Open-Systems-Pharmacology/docs/pull/339)
- The lipophilicity constraint (MoBi #2445) is accurately cited as a v13 migration blocker. [[github](https://github.com/Open-Systems-Pharmacology/docs/pull/339)](https://github.com/Open-Systems-Pharmacology/docs/pull/339)

Given that, the remaining issues are mostly:

- Making the migration workflow **fully aligned** with the building-block list it claims to cover (add spatial structures and more explicit event/app guidance).  
- Cleaning up CI and lint findings (dictionary and fenced code language).  
- Ensuring no leftover text suggests a molecule-type behavioral difference across v12 vs v13.

Once those are fixed, the PR looks technically solid and well-scoped.
</details>

part-4/converting-v12-projects-to-v13.md
```diff
@@ -0,0 +1,233 @@
+# Converting v12 projects to v13
+
+Version 13 of the OSP Suite refines the [modularization concept](modularization-concept.md) introduced in v12. The organizational structure (Modules → Building Blocks → Simulations) is unchanged, and the *parameter value* and *initial condition* resolution orders are unchanged. What changed is the **merge behavior of several building-block types when modules are combined** — specifically the difference between the **"Extend"** and **"Overwrite"** merge modes.
```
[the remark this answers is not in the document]
PavelBal: Partially valid — fixed in 7691ac7, with one correction to the attribution.

Of the words named, only **`micellar`, `instantiations`, `redefinitions` and `unioned`** are actually introduced by this page. **`modularization`/`Modularization` and `BBs` are pre-existing**: they already fail on files this branch does not touch — `part-4/first-steps.md`, `example-workflows.md`, `spatial-structures-bb.md`, `expression-profiles-bb.md`, `individuals-bb.md`, `initial-conditions-bb.md`, `parameter-values-bb.md` and `SUMMARY.md`.

What I changed:

- Added `modularization`, `Modularization`, `BBs`, `micellar`, `instantiations`, `redefinitions` to `.github/workflows/wordlist.txt`. I verified this is the right file: `Config/.spellcheck.yml` in the `Workflows` repo lists `.github/workflows/wordlist.txt` and `./wordlist_osp_global.txt` as the two aspell wordlists.
- Replaced `unioned` with "combined into the union of both definitions" — agreed that it is unnecessary jargon.

**The Spellcheck job will still be red after this**, and not because of this PR. Roughly 50 further words fail across files this branch never touches: the SI-unit symbols in `appendix.md` (`Bq`, `Gy`, `Sv`, `kat`, `lx`, `sr`, `cd`, …), LaTeX fragments in `part-3` (`cdot`, `mathrm`, `dt`, `dA`, `Ic`, `Iu`), plus `extrahepatic`, `reusability`, `lognormal`, `unary`, `PTs`, `renamings` and others. The same job fails identically on #337, #340 and #341, which are one-line changes. Whitelisting those (or fixing the LaTeX ones at the source) belongs in a separate PR, or in the global wordlist in the `Workflows` repo — it should not be bundled here.

part-4/converting-v12-projects-to-v13.md
```diff
@@ -0,0 +1,233 @@
...
+
+{% hint style="danger" %}
+**Combining PK-Sim modules set to "Overwrite" can fail to build.** In v12, PK-Sim modules were imported with the default merge behavior **"Overwrite"**, and combining them still worked: even under "Overwrite", the molecule **include/exclude lists** of passive transports and observers were extended rather than replaced. In v13, "Overwrite" replaces those lists too, so a passive transport keeps only the molecules of the module that overwrites it.
+
+This surfaces when combining **two large-molecule models**. The FcRn-mediated `NetMassTransfer_*` transports apply to their own module's molecules via those molecule lists; once the lists are replaced instead of combined, the molecule-specific entities the transport formulas reference are no longer created and **simulation creation fails** with errors such as:
+
+```
+Transport 'NetMassTransfer_InterstitialToEndosomal' references an entity with path
+'<Molecule>-FcRn_Complex|Is small molecule' that cannot be found
+```
```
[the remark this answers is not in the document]
PavelBal: Fixed in 7691ac7 — the fence is now ```text.

part-4/converting-v12-projects-to-v13.md
```diff
@@ -0,0 +1,233 @@
...
+
+   Exported simulations may contain `SimulationEntitySource` provenance records (each entity traced to its origin `moduleName`, `buildingBlockType`, `sourcePath`). These do **not** affect model behavior. Their presence and count depend on the MoBi/export build rather than on model semantics — the same OSP version can export with or without them — so **do not treat them as a v12↔v13 difference; ignore them when comparing.**
+   {% endhint %}
+
+6. **Decide per module: Extend or Overwrite.**
+   - If the module was meant to **fully replace** a molecule/reaction/transport/observer (the common v12 assumption), set its merge mode to **"Overwrite"**.
+   - If it was meant to **add to** an existing building block, keep **"Extend"** and verify the merged result — remove any now-redundant duplicated content, and remember that reaction educts/products cannot be removed (use stoichiometry `0`).
```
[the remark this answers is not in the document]
PavelBal: Mostly valid — fixed in 7691ac7, with one part of the suggestion corrected.

**Spatial structures: correct, real gap.** The page's own table flags `MoleculeProperties` and neighborhoods as changed, but steps 4 and 6 never told the reader to look at them. Both steps now do.

**Events: already covered in step 4** before this change — the sentence beginning "Also flag any event/application defined with the same name in more than one module … verify the administered molecule in the built simulation". I have kept that content, promoted it to its own bullet so it is not buried at the end of a paragraph, and added the explicit merge decision to step 6.

**Pushing back on one point:** the suggestion to "indicate when you want `MoleculeProperties` merged vs replaced" describes a choice that does not exist. Per the Spatial structure section, `MoleculeProperties` are extended in **both** modes in v13, so the merge mode cannot control them at all — a v12 module that replaced them wholesale cannot be restored to that behavior by setting "Overwrite". Step 6 now states this explicitly and directs the reader to change the module *content* instead. Neighborhoods genuinely are mode-sensitive (replaced under "Extend", overwritten under "Overwrite"), but under neither mode can a neighborhood be removed, and an invalid redefinition silently keeps the earlier one — that is stated too.

While in these steps I also tightened the scope to **every module after the first** rather than "each Extension module", which lines up with the multi-module nuance corrected earlier in this PR.

PavelBal:
@Yuri05 Checked both. Fixes are in 7691ac7; individual replies are in the three CodeRabbit threads. Summary, including one point where I disagreed with the suggestion.

## The three red checks are not caused by this PR

This is the most important finding, since it looks like this PR is blocking merge on CI. It is not:

| Check | This page's contribution |
|---|---|
| `markdown-link-check` | **0** dead links (16 checked in this file). The 5 + 3 + … failures are in `references.md` (dead FDA/ScienceDirect URLs) and files linking to the missing `part-4/model-building-components.md`. |
| `crossref-check-without-anchors` | **0** of the 57 invalid references. |
| `Spellcheck` | **0** words, after 7691ac7. |

All three fail identically on **#337, #340 and #341**, which are one-line changes, so they are pre-existing repo-wide conditions on `v13`. #341 already fixes the crossref job.

After 7691ac7 the spellcheck job is still red, but only on 44 words in 12 files this branch does not touch: SI-unit symbols in `appendix.md`, LaTeX fragments in `part-3` (`cdot`, `mathrm`, `dt`, `dA`) and `part-4/parameters-formulas-tags.md`, plus `extrahepatic`, `reusability`, `lognormal`, `unary`, `PTs`, `renamings`, `StartEvent`, `subnodes`, `nondisintegrating`. My wordlist additions did incidentally clear ~20 pre-existing failures (`modularization`, `Modularization`, `BBs`) across eight other pages. I deliberately did **not** whitelist the rest here: the LaTeX ones are a source bug (those formulas should be fenced, not spell-exempted), and the rest belong in a cleanup PR or in the global wordlist in the `Workflows` repo. Two entries that already exist in the wordlist — `Türk’s` and `Yun’s` — are still flagged in `part-7/application-simulation.md`, which looks like a curly-apostrophe mismatch worth a separate look.

## Perplexity's points

1. **Residual molecule-type claims — checked, none found, no change needed.** The only remaining mentions are `Molecule **type** … are overwritten — the later module wins, **as in v12**` and "type, stationary, calculation methods and parameter type are overwritten by the later module in both modes, so they **need no migration attention**". The Molecules row of the summary table makes no claim about type at all. So nothing implies a v12↔v13 type difference.

2. **Spatial structures and events in the checklist — fixed.** Spatial structures were a genuine gap and are now in both step 4 and step 6. Events were already in step 4; that content is kept, promoted to its own bullet, and given an explicit decision bullet in step 6. **I did not follow the suggestion to describe choosing whether `MoleculeProperties` are "merged vs replaced"** — that choice does not exist. They are extended in *both* modes in v13, so no merge-mode setting restores the v12 wholesale-replacement behavior; step 6 now says so and points the reader at the module content instead. Neighborhoods are mode-sensitive, but can never be *removed*, and an invalid redefinition silently keeps the earlier one — also stated.

3. **Spellcheck** — see above. `unioned` is now "combined into the union of both definitions".

4. **MD040** — the error-output fence is now ` ```text `.

5. **Multi-module nuance in the checklist — fixed.** Step 4 now scans "every module after the first" instead of "each Extension module", and step 6 opens with "The first module in the hierarchy has nothing to merge into, so its mode has no effect — make this decision for every module selected after it, not only the last one." Nothing in the page claims that setting all PK-Sim modules to "Extend" is universally harmless; the "Extend" recommendation is scoped to the specific PK-Sim-module combination failure.

6. **Overall soundness** — no action; matches the audit above.

Yuri05 (approved):

PR #343, opened 2026-07-31
Correct cumulated sensitivity formula in Sensitivity Analysis

### LINKED ISSUE #71: Description of cumulated sensitivity is manual is wrong
The formula defined in the manual is not the one used in code!
image

whereas the code uses simple Sum of absolute values
https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/blob/f75c2b2f6a639a8a8f27e4937358ad7819e87ba0/src/OSPSuite.Core/Domain/SensitivityAnalyses/SensitivityAnalysis.cs#L128

### LINKED ISSUE #73: Sensitivity Analysis Description misleading
The sensitivity analysis in PK Sim allows a visualization for a summary of the relevant Parameters accounting for 90% of the cumulated sensitivity. The Manual describes the computation applying the sqrt to both, the cumulated sensitivities as well as to the total sensitivity in the denominator.
The plot actually visualizes parameters derived when neglecting the sqrt. Also, the formula in the manual is not considering absolute values

PavelBal:
Fixes #73
Fixes #71

Issues #71 and #73 are duplicates: both report that the cumulated sensitivity formula in the manual does not match the implementation. #73 (JanSchlender, March 2017) emphasises the spurious square root and the missing absolute values; #71 (msevestre, September 2017) states that "the code uses simple Sum of absolute values". Same formula, same section, one fix.

## Problem

The **Tab Sensitivity Analysis** section of the manual documented the cumulated sensitivity used to pick the parameters shown in the ranking chart as

$$S_{Cum,l,j} = \sqrt{\sum_{i=1}^{l} S_{i,j}} \big/ \sqrt{\sum_{i=1}^{nP} S_{i,j}}$$

Two things were wrong, both raised in the issue: the square roots are not applied by the implementation, and the sensitivities are summed as signed values rather than absolute values.

## What the implementation does

Verified against `sensitivitiesUpToTotalSensitivity` in [`SensitivityAnalysisRunResult.cs`](https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/blob/develop/src/OSPSuite.Core/Domain/SensitivityAnalyses/SensitivityAnalysisRunResult.cs) (the line referenced in the issue discussion has since moved):

- sensitivities are ordered descending by absolute value via `OrderByDescending(x => Math.Abs(x.Value))`;
- the total is `orderedSensitivities.Sum(x => Math.Abs(x.Value))` — absolute values, no square root;
- the running sum also accumulates absolute values, and `TakeWhile` evaluates `runningSensitivity / totalSensitivity >= threshold` *before* adding the current element, so the parameter that crosses the threshold is still included;
- `NaN` sensitivities are filtered out beforehand, so `nP` counts only parameters with a defined sensitivity;
- the threshold is the fixed constant `TOTAL_SENSITIVITY_THRESHOLD = 0.9` in `SensitivityAnalysisPKParameterAnalysisPresenter`.

## Change

The formula was a raster image (`assets/images/part-5/scum-lj.jpg`), so it could not be corrected in place. It is replaced with an inline GitBook LaTeX block, consistent with the math already used in [part-3/pk-sim-compounds-defining-inhibition-induction-processes.md](../blob/v13/part-3/pk-sim-compounds-defining-inhibition-induction-processes.md), which also makes the formula editable in future.

Substantive corrections to the surrounding text: the square roots are removed, absolute values are applied to each sensitivity before summing, the sort is stated explicitly as *descending*, `nP` is qualified as counting only parameters with a defined sensitivity, and the cut-off is given as a precise condition (smallest `l` with `S_Cum,l,j >= 0.9`) instead of the previous garbled sentence.

The obsolete image is deleted; line 157 of the markdown was its only reference anywhere in the repository.

## Why the manual and not the code

@Yuri05 asked in #71 whether the manual or the code should be corrected. This PR corrects the manual, for the reasons set out in [a comment on #71](https://github.com/Open-Systems-Pharmacology/docs/issues/71) — in short: the square root has never existed anywhere in the OSPSuite.Core history; the formula as drawn is ill-defined for negative sensitivities and so was never implementable; Euclidean aggregation has no interpretation for local first-order sensitivities; and adopting it would shrink the displayed parameter set drastically (in a simulation of 200,000 random sensitivity vectors it never showed *more* parameters and on average showed 6.5 fewer), silently altering every previously published sensitivity plot in the GUI, `ospsuite` and the Reporting Engine, which all share this one code path.

## Checks

- Repo-wide grep confirms no remaining reference to `scum-lj`, so the link checker has nothing dangling.
- The LaTeX uses only `\frac`, `\sum`, `\left|`, `\dots` and `\geq`, all of which already appear in part-3 and pass the spellcheck workflow, so no `wordlist.txt` additions were needed.

## Out of scope

The four other formula images in this file (`sij.png`, `sij-e.png`, `1-a-k-n.png`, `1-1-a-kn.png`) appear mathematically correct and were left untouched; converting them to LaTeX would be unrelated cleanup.


Yuri05:
>Two things were wrong, both raised in the issue: the square roots are not applied by the implementation, and the sensitivities are summed as signed values rather than absolute values.

the 2nd statement is clearly wrong: the sensitivities are summed as **absolute values** (s. [here](https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/blob/develop/src/OSPSuite.Core/Domain/SensitivityAnalyses/SensitivityAnalysisRunResult.cs#L109)).

And regarding the first part: I think we should rather fix the **implementation** (replace $\frac{runningSensitivity}{totalSensitivity}$ with $\sqrt\frac{runningSensitivity}{totalSensitivity}$ [here](https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/blob/develop/src/OSPSuite.Core/Domain/SensitivityAnalyses/SensitivityAnalysisRunResult.cs#L106)) and not the docu, no? @katrincoboeken

Yuri05 (changes requested):
s. https://github.com/Open-Systems-Pharmacology/docs/pull/343#issuecomment-5179152436

KatrinCoboeken:
For me the most logical solution seems to be this formula: 
$$Scum_{l,j} = \frac{\sqrt{\sum_{i=1}^l(S_i,j)^2}}{\sqrt{\sum_{i=1}^{np}(S_i,j)^2}}$$

We used this also  in reports generated with Matlab before sensitivity was implemented in PK-Sim.
So my suggestion would be adjust both, implementation and manual

Yuri05:
> For me the most logical solution seems to be this formula: S c u m l , j = ∑ i = 1 l ( S i , j ) 2 ∑ i = 1 n p ( S i , j ) 2
> 
> We used this also in reports generated with Matlab before sensitivity was implemented in PK-Sim. So my suggestion would be adjust both, implementation and manual

Issue created: https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/issues/2923

PR #345, opened 2026-07-31
Document the objective function and Total Error of Parameter Identification

### LINKED ISSUE #74: More details on how "Total Error" in PI is calculated
We have often received questions how the "Total Error" in the PI is calculated for each Algorithm.

Maybe we can add the explicit formula for each algorithm.

PavelBal:
Closes #74.

The manual mentioned the total error only as a value displayed in the tab **Results** and described the residuals informally in the Scaling table. The recurring question of how the Total Error is calculated — and whether it differs per optimization algorithm — was not answered anywhere in the documentation.

### What was added

A new section **Objective function and Total Error** at the end of *Configuration of Optimization* (after the LLOQ handling, right before the algorithm descriptions), containing:

* the per-point residual $r_{k,i} = w_k \times w_{k,i} \times (g(s_{k,i}) - g(o_{k,i}))$ with the mapping weight and the point weight,
* the scaling transformation $g$, including the base of the logarithm and the $\epsilon = 10^{-20}$ clamping,
* the LLOQ transformation applied before the scaling, including the note that the LLOQ/2 of the legacy mode comes from the import of the observed data rather than from the residual calculation,
* the four conditions under which an observed data point does not contribute at all,
* the Total Error $E_{total} = \sqrt{\sum_k \sum_i r_{k,i}^2}$,
* a statement that all three algorithms minimize this same objective (Monte-Carlo and Nelder-Mead evaluate it directly, Levenberg-Marquardt minimizes the sum of squares of the same residual vector), and a hint that the unnormalized sum is only comparable between fits using the same data, weights and scaling.

### Supporting edits in the same file

* The Background now points at the new section instead of at the Scaling table.
* The Scaling table states the base of the logarithm explicitly ($\log_{10}$), and a sentence after the table notes that both weights multiply the difference.
* The `eps=10E-20` notation in the LLOQ warning is written as $\epsilon = 10^{-20}$.
* The *Residuals vs. Time* description identifies the plotted values as the $r_{k,i}$ defined in the new section.
* `sqrt` added to the spellcheck wordlist.

### Verification

All formulas were checked against the current implementation in `OSPSuite.Core` (`b941097d`): `ResidualCalculator.calculateOutputResiduals`, `ResidualCalculatorForOnlyObservedData`, `ResidualCalculatorForSimulationOutputAsObservedDataLLOQ`, `TimeGridRestrictor`, `ResidualsResult.TotalError`, `MPFitLevenbergMarquardtOptimizer`, `MonteCarloOptimizer` and `NelderMeadOptimizer`.

## Summary by CodeRabbit

- **Documentation**
  - Expanded Parameter Identification guidance with clear definitions for residuals, total error, transformations, weighting, LLOQ handling, and skipped data points.
  - Clarified base-10 log scaling, epsilon handling, objective functions, algorithm equivalence, and residual plot interpretation.

- **Chores**
  - Added `sqrt` to the wordlist.


Yuri05:
@coderabbitai full review

part-5/parameter-identification.md
```diff
@@ -228,6 +230,45 @@ If in Observed Data LLOQ values are contained as 0 values, for **Remove data bel
...
+
+For an output mapping $$k$$ and an observed data point $$i$$ measured at the time $$t_{k,i}$$, the residual is
+
+$$
+r_{k,i} = w_k \times w_{k,i} \times (g(s_{k,i}) - g(o_{k,i}))
+$$
+
+where $$o_{k,i}$$ is the observed value, $$s_{k,i}$$ is the value of the mapped simulation output at $$t_{k,i}$$, $$w_k$$ is the weight of the mapping and $$w_{k,i}$$ is the weight of the individual data point (see [Mapping Simulation Outputs to Observed Data](#mapping-simulation-outputs-to-observed-data)). Observed and simulated values are converted to the internal base units before the residual is calculated.
+
+The transformation $$g$$ is given by the **Scaling** of the mapping: $$g(y) = y$$ for Lin scaling and $$g(y) = \log_{10}(y)$$ for Log scaling, where values smaller than $$\epsilon = 10^{-20}$$ are replaced by $$\epsilon$$ before the logarithm is applied. Since observed values below $$\epsilon$$ are excluded from a log scaled output (see below), this replacement applies to simulated values only.
```
Yuri05, line 247: _...Since observed values below $$\epsilon$$ are excluded from a log scaled output (see below),..._
Even if no LLOQ was defined for a dataset? Hmm, never knew this...

part-5/parameter-identification.md
```diff
@@ -228,6 +230,45 @@ If in Observed Data LLOQ values are contained as 0 values, for **Remove data bel
...
+
+All optimization algorithms minimize the same objective function, the **Total Error**. It is the square root of the sum of the squared residuals over all output mappings and it is the value that is reported in the tab **Results** and plotted in the error history of the **Visual Feedback**.
+
+For an output mapping $$k$$ and an observed data point $$i$$ measured at the time $$t_{k,i}$$, the residual is
+
+$$
+r_{k,i} = w_k \times w_{k,i} \times (g(s_{k,i}) - g(o_{k,i}))
+$$
+
+where $$o_{k,i}$$ is the observed value, $$s_{k,i}$$ is the value of the mapped simulation output at $$t_{k,i}$$, $$w_k$$ is the weight of the mapping and $$w_{k,i}$$ is the weight of the individual data point (see [Mapping Simulation Outputs to Observed Data](#mapping-simulation-outputs-to-observed-data)). Observed and simulated values are converted to the internal base units before the residual is calculated.
```
Yuri05, line 243: Would also add that $$o_{k,i}$$ is always the **mean** value of an observed data point - **SD is ignored** in the optimization, whether defined or not.

part-5/parameter-identification.md
PavelBal: Yes — it is independent of LLOQ. I checked against the implementation, because the current wording does not make that obvious.

The filter is in `TimeGridRestrictor`:

```csharp
private static bool valueIsValid(Scalings scaling, float value)
{
   return scaling == Scalings.Linear || value >= Constants.LOG_SAFE_EPSILON;
}
```

No LLOQ enters that predicate — only the scaling and the magnitude of the value. It is reached from all three `RemoveLLOQMode` branches, including `getRelevantIndicesAll`, which is the one used when the mode is `Never`. So a log scaled mapping drops those points even for a data set with no LLOQ defined at all. `Constants.LOG_SAFE_EPSILON` is `1e-20F`, the $$\epsilon$$ of the text, and the interface docstring states the intent outright: *"The `outputScaling` is used to remove possible 0 values when the output is using a `Log` scaling."*

Since the test is `value >= ε`, what this removes in practice is observed values that are **zero or negative** — not values that merely happen to be very small. That is the part I think was genuinely unclear, so I am making it explicit.

Worth noting that the sentence right after it depends on this filter. Both residual calculators apply `LogSafe` to *both* arguments:

```csharp
return LogSafe(simulatedValue) - LogSafe(observedValue);
```

so "this replacement applies to simulated values only" holds only *because* sub-$$\epsilon$$ observed values were already excluded upstream. The two statements are load-bearing on each other, which is an argument for stating the exclusion clearly rather than in a subordinate clause.

I have prepared the following clarification (not yet pushed):

> The exclusion of observed values below $$\epsilon$$ is independent of LLOQ. It applies to every log scaled mapping, also when no LLOQ is defined for the data set, and in practice it removes observed values that are zero or negative, which have no logarithm.

and extended the skip-list bullet with "regardless of whether an LLOQ is defined."

Yuri05: ok, interesting.

Yuri05 (approved):

PR #346, opened 2026-07-31
Fix outdated guidance on time-based event conditions

### LINKED ISSUE #111: Events start condition - use ">=" instead of "="?
Application events conditions are defined as 

`Time = StartTime`

and so on. The documentation states

> The easiest way to let an event happen at a given simulation time would now be to enter the formula "Time > 500", which would execute the event at 500 minutes. The use of "> 500" instead of "= 500" is advantageous since it might well be that during the simulation, the exact value of 500 will never be assumed, depending on the time step. If you plan to quickly test different values for this time, it is advantageous to define this execution time as a parameter which can be altered in the simulation.

Shouldn't the default events' conditions be changed according to this suggestion?

PavelBal:
Closes #111.

## Problem

The Events building block documentation recommended `Time > 500` over `Time = 500` for the start condition of an event, with the rationale:

> The use of "> 500" instead of "= 500" is advantageous since it might well be that during the simulation, the exact value of 500 will never be assumed, depending on the time step.

That rationale is wrong, as explained by @Yuri05 in the issue. If a condition compares `Time` with an expression that is constant during the simulation, SimModel adds the value of that expression to the solver time points, so the condition *is* evaluated at exactly that time - independently of the output time raster. Using `>` is in fact the worst choice: the condition is evaluated at 500, returns `false` there, and the event is only executed at the next output time point, which makes the execution time depend on the output resolution.

I re-checked this against the current SimModel source before writing the text: [`BooleanFormula::SwitchTimePointFromComparisonFormula`](https://github.com/Open-Systems-Pharmacology/OSPSuite.SimModel/blob/master/src/OSPSuite.SimModelNative/src/BooleanFormula.cpp#L149-L187) handles both operand orders (`Time OP <formula>` and `<formula> OP Time`), registers the switch time point only if the other operand can be evaluated without ODE variables, and is used by all comparison formulas (`=`, `>=`, `>`, `<=`, `<`, `!=`), while `AND`/`OR`/`NOT` collect the time points of their operands recursively.

## Changes

All in [part-4/events-bb.md](part-4/events-bb.md):

- **Event conditions**: new paragraph documenting that a comparison of `Time` with an expression constant during the simulation is guaranteed to be evaluated at exactly that time, including the `Time = 10 OR Time = 20 OR Time = P1 OR Time = P1 + P2` example, and a warning against using `>` for the start time.
- The existing warning about conditions depending on the system state now opens with the contrast: for state-dependent conditions no time point can be determined in advance, so only the output time points are used.
- **Tutorial**: steps 6 and 13 now use `Time = 500` / `Time = E1Time`. The "define the time as a parameter" tip was kept and moved into the general section as well.
- The paragraph about a **permanently active** event now points out that this requires `Time >= E1Time`; with `Time = E1Time`, unchecking **One Time Event** has no effect because the condition is only `true` at a single time point. This was implicitly wrong before as well, since the old text combined `Time > 500` with "keeping the amount constant after 500 minutes".
- The example condition in the **One time** warning was changed from `TIME > StartTimeParameters` to `TIME >= StartTimeParameter` (also fixes the plural typo).

## Notes for the reviewer

- The screenshot [event-building-completed.jpg](assets/images/part-4/event-building-completed.jpg) still shows `Time > E1_Time` in the condition input box and therefore no longer matches step 13. It needs to be retaken in MoBi - I did not want to edit the image itself. (The same screenshot also uses the alias `E1_Time` while the text says `E1Time`, which was already inconsistent before this PR.)
- `master` and `v12` contain the same outdated passage. This PR only targets `v13` as requested; let me know if the fix should be forward-ported to `master`.


PavelBal:
Note on the two red checks - both fail on files this PR does not touch and already failed the same way on `v13` before (e.g. #339):

- **Spellcheck**: `reusability` in `part-4/building-block-concepts.md` and `nondisintegrating` / `cdot` / `dA` / `dt` in `part-3/pk-sim-formulations.md`. `part-4/events-bb.md` itself is processed without findings.
- **markdown-link-check**: 6 dead external links, all in `references.md` (a DOI, three ScienceDirect entries and two FDA pages). For `part-4/events-bb.md` the report says "19 links checked" with no failures.

PavelBal:
I traced the full mechanism through the current SimModel `master` to make sure the new text is accurate, not just plausible:

| Step | Code | Behavior |
| --- | --- | --- |
| Register | [`BooleanFormula::SwitchTimePointFromComparisonFormula`](https://github.com/Open-Systems-Pharmacology/OSPSuite.SimModel/blob/master/src/OSPSuite.SimModelNative/src/BooleanFormula.cpp#L149-L187) | detects `Time OP f` / `f OP Time` via `IsTime()` and evaluates the other operand as `f->DE_Compute(NULL, 0.0, ...)`. The `NULL` y-vector is what implements "constant during the simulation": any dependence on ODE variables throws and the time point is silently dropped. |
| Operators | `EqualFormula`, `GreaterEqualFormula`, `GreaterFormula`, `LessEqualFormula`, `LessFormula`, `UnequalFormula` | all six override `SwitchTimePoints()`. `AND`/`OR`/`NOT` use the base `BooleanFormula::SwitchTimePoints()`, which concatenates the time points of both operands - hence the `OR` example works. |
| Collect | `Switch::SwitchTimePoints` → `SwitchTask::SwitchTimePoints` | over all switches of the simulation. |
| Merge | [`SimulationTask::OutputTimePoints`](https://github.com/Open-Systems-Pharmacology/OSPSuite.SimModel/blob/master/src/OSPSuite.SimModelNative/src/SimulationTask.cpp#L114-L145) | a switch time point that is not a user output time point is inserted with `saveSolution = false`, `isSwitchTimePoint = true`. |
| Stop exactly | `PerformSolverStep(outTimePoint.Time(), ...)` with `assert(solverOutputTime == outTimePoint.Time())` on success | the solver lands on exactly that value. |
| Evaluate | `Simulation::PerformSwitchUpdate(solution, solverOutputTime)` ([DESolver.cpp#L367](https://github.com/Open-Systems-Pharmacology/OSPSuite.SimModel/blob/master/src/OSPSuite.SimModelNative/src/DESolver.cpp#L367)) | called unconditionally for every entry of the merged list, not only for switch time points. |
| One time | `Switch::PerformSwitchUpdate`: `if (_oneTime && _wasFired) return false;` | prevents re-firing. |

`EqualFormula::DE_Compute` uses `std::equal_to<>`, i.e. an exact floating point comparison (the source even flags it as "potentially unsafe"). It is nevertheless reliable in this case, because the value compared against is the very same double that was registered as the switch time point: `TIME` returns the stop time, the other operand recomputes the same constant formula, and the comparison is `V == V`. This is also why the recommendation is specific to expressions that are constant during the simulation.

One correction to the issue discussion: [this comment](https://github.com/Open-Systems-Pharmacology/docs/issues/111#issuecomment-676696271) assumed that a start time of 25 with the output schema 10/30/50/70 yields the outputs 10/25/30/50/70. In SimModel that is not the case - the switch time point is inserted with `saveSolution = false`, so it is a solver stop but not a stored result. The event still fires at exactly 25; only the stored outputs are unaffected. (PK-Sim may add application times to the output schema by itself, which is presumably where the recollection came from.) I added a short info hint covering this, since a user whose output grid does not contain the event time would otherwise see the change appear at the next output point and conclude that the event fired late - exactly the misconception this PR removes.

Yuri05 (approved):

PR #312, opened 2025-11-04
- Simulations rework

PavelBal:
Fixes https://github.com/Open-Systems-Pharmacology/docs/issues/284

PR #350, opened 2026-08-04
Merge v12 into v13

PavelBal:
Brings the four commits that landed on `v12` but never reached `v13`:

- `9806162` — Simulations rework (#312)
- `ab97d62` — GitHub Action to detect non-printable characters (#320)
- `15481ee` — Remove zero-width non-joiner (ZWNJ, `U+200C`) characters (#321), including three file renames
- `458719f` — FcRn concentration default species: monkey → human (#325)

`v13` currently still contains ZWNJ characters in 44 Markdown files, so the cleanup from #321 is genuinely needed here.

## Conflicts

The merge conflicts in 11 files / 26 hunks. Almost all of them are the *same repair made twice*: both branches independently fixed the links that dangled after `model-building-components.md` was split into per-building-block pages. In those cases `v13`'s version is the better one, so **take `v13` (`--ours`) for the hunk and then re-strip the ZWNJ characters** — note that plain `git checkout --ours <file>` is wrong, because it throws away the ZWNJ cleanup in the non-conflicting parts of the same file.

| File | Hunks | Resolution |
| --- | --- | --- |
| `part-1/principles-of-pbpk-modeling.md` | 1 | v13 — anchor `#distribution---distribution-calculation` is the precise target |
| `part-4/building-block-concepts.md` | 1 | v13 — equivalent rewording |
| `part-4/events-bb.md` | 3 | **v13 — important.** v12 still says `Time > 500`; #346 deliberately corrected this to `Time = 500` and added the "Conditions based on the simulation time" section. Taking v12 here would revert that fix. |
| `part-4/first-steps.md` | 2 | v13 — keeps the pointer to `building-block-concepts.md` and the fixed link text |
| `part-4/molecules-bb.md` | 5 | v13 — v12's variants contain two stray `))` typos |
| `part-4/parameters-formulas-tags.md` | 2 | v13 — v12 drops the Observers/Events links |
| `part-4/passive-transports-bb.md` | 2 | v13 — v12 points *Molecule Parameters* at `#example---creating-new-molecules` and has a stray `))` |
| `part-4/reactions-bb.md` | 2 | v13 — same fix, better link text |
| `part-4/reuse-of-project-information-from-previous-versions.md` | 1 | v13 — the page was rewritten as "Project conversion workflows" (#339); v12 only stripped a ZWNJ from the old heading |
| `part-4/spatial-structures-bb.md` | 2 | v13 — keeps the corrected `ProjectNew.svg` icon reference |
| `part-4/setting-up-simulation.md` | 5 | **v12 (`--theirs`), whole file.** This is the #312 rework and `v13` does not have it. v13's only changes to this file are link fixes, all of which v12 already makes correctly (`#send-to-mobi`, `../part-5/parameter-identification.md`, and the scale-divisor hint moved into *ODE Solver Properties* pointing at `#calculate-scale-divisors`). Keeping v13 here would also leave references to `SimCW-MSV.png`, `SimCW-PSV.png` and friends, which #312 deletes. |

## Verification of the proposed resolution

Applied locally against this merge:

- no conflict markers left
- 0 files containing `U+200C` (also checked `U+200B`, `U+200D`, `U+FEFF`) — the new `NonPrintableCheck` action should pass
- 0 missing image/asset references
- 0 broken internal Markdown links or heading anchors

## One thing to look at that did *not* conflict

`SUMMARY.md` auto-merged to v12's navigation order, which moves **Parameters, Formulas, Tags, and Keywords** from just after *Initial Conditions* to just before *Simulations*, and renames the entry *Setting up a Simulation* → *Simulations*. The rename follows #312 and is correct; the reordering is a silent side effect worth a conscious decision.


PavelBal:
Superseded by #351, which carries the same merge with conflicts resolved.

This PR cannot be made mergeable in place: resolving a PR's conflicts means pushing the resolution onto the head branch, and that would pull all 38 `v13` commits into the released `v12` branch. #351 puts the merge commit on a topic branch off `v13` instead, leaving `v12` untouched.

PR #347, opened 2026-07-31
Add a "How to Cite this Manual" page

PavelBal:
Fixes #201

Adds a top-level page, listed in `SUMMARY.md` right after the copyright page, that tells readers how to cite the manual.

## What is in it

- A plain-text citation and a BibTeX `@Manual` entry, following the format @Yuri05 proposed in the issue.
- A note that the `edition`, `year`, and `month` must be taken from the copyright page of the version the reader actually used — those two pages have to be bumped together at each release.
- A short section pointing at [Lippert et al. (2019)](https://doi.org/10.1002/psp4.12473) for citing the *software platform*, which is what people usually want when they reach for a citation, rather than the manual.

## One deviation from the proposed snippet

The `author` value is wrapped in braces:

```bibtex
author = "{Open Systems Pharmacology Community}",
```

Without them BibTeX parses "Open Systems Pharmacology" as given names and "Community" as a surname, and most styles render the author as "O. S. P. Community". The page explains this so the braces do not get "cleaned up" later.

## Notes

- Version and date currently match `copyright.md` on this branch (Version 12 Update 1, September 2025). They will need bumping along with `copyright.md` when v13 is released.
- `BibTeX`/`bibtex` added to the spellchecker wordlist.
- The entry was verified to parse with `bibtexparser`, and all relative links in the new and changed files resolve.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Documentation**
  * Updated the handbook introduction to describe its examples, references, and PBPK model development focus.
  * Documented Manual Version 13 (2026), compatibility boundaries, publication details, and updated copyright information.
  * Added a citation guide with plain-text and BibTeX formats, including guidance for citing earlier versions and related publications.
  * Added the citation guide to the table of contents.

PR #351, opened 2026-08-04
Merge v12 into v13 (conflicts resolved)

PavelBal:
Merges `v12` into `v13` with conflicts resolved. Replaces #350, which could not be made mergeable in place — resolving a PR's conflicts means pushing the resolution onto the head branch, and here that would have pulled all 38 `v13` commits into the released `v12` branch. This branch carries the merge commit instead, so `v12` is untouched.

Brings the four commits that landed on `v12` but never reached `v13`:

- `9806162` — Simulations rework (#312)
- `ab97d62` — GitHub Action to detect non-printable characters (#320)
- `15481ee` — Remove zero-width non-joiner (ZWNJ, `U+200C`) characters (#321), including three file renames
- `458719f` — FcRn concentration default species: monkey → human (#325)

`v13` still contained ZWNJ characters in 44 Markdown files, so the cleanup from #321 was genuinely needed here — and without it the new check from #320 would have failed on `v13`.

## How the conflicts were resolved

11 files / 26 hunks conflicted. Almost all of them are the *same repair made twice*: both branches independently fixed the links that dangled after `model-building-components.md` was split into per-building-block pages. In those cases `v13`'s version is the more accurate one, so the hunk keeps `v13` and the ZWNJ strip is re-applied on top. (Plain `git checkout --ours <file>` would have been wrong here — it restores the whole file from `v13` and so silently discards the ZWNJ cleanup in the file's non-conflicting parts.)

| File | Hunks | Resolution |
| --- | --- | --- |
| `part-1/principles-of-pbpk-modeling.md` | 1 | v13 — `#distribution---distribution-calculation` is the precise anchor |
| `part-4/building-block-concepts.md` | 1 | v13 — equivalent rewording |
| `part-4/events-bb.md` | 3 | **v13 — important.** v12 still says `Time > 500`; #346 deliberately corrected this to `Time = 500` and added the "Conditions based on the simulation time" section. Taking v12 would have reverted that fix. |
| `part-4/first-steps.md` | 2 | v13 — keeps the pointer to `building-block-concepts.md` and the fixed link text |
| `part-4/molecules-bb.md` | 5 | v13 — v12's variants contain two stray `))` typos |
| `part-4/parameters-formulas-tags.md` | 2 | v13 — v12 drops the Observers and Events links |
| `part-4/passive-transports-bb.md` | 2 | v13 — v12 points *Molecule Parameters* at `#example---creating-new-molecules` and has a stray `))` |
| `part-4/reactions-bb.md` | 2 | v13 — same fix, better link text |
| `part-4/reuse-of-project-information-from-previous-versions.md` | 1 | v13 — the page was rewritten as "Project conversion workflows" (#339); v12 only stripped a ZWNJ from the old heading |
| `part-4/spatial-structures-bb.md` | 2 | v13 — keeps the corrected `ProjectNew.svg` icon reference |
| `part-4/setting-up-simulation.md` | 5 | **v12, whole file.** This is the #312 rework and `v13` does not have it. v13's only changes to this file are link fixes, all of which v12 already makes correctly (`#send-to-mobi`, `../part-5/parameter-identification.md`, and the scale-divisor hint moved into *ODE Solver Properties* pointing at `#calculate-scale-divisors`). Keeping v13 here would also have left references to `SimCW-MSV.png`, `SimCW-PSV.png` and others that #312 deletes. |

## Verification

Run against the merged tree:

- no conflict markers remain
- the check from #320 passes: 0 filenames and 0 Markdown lines with non-printable characters, using that workflow's exact logic over the whole repo (`U+200B`–`U+200F`, `U+FEFF`, and the control ranges)
- 0 missing image or asset references
- 0 broken internal Markdown links or heading anchors

## About the two red checks

**Spellcheck** and **markdown-link-check** fail on every `v13`-based PR right now, including the already-merged #347 — they are pre-existing and branch-wide, not caused by this merge. I diffed both against #347 as the baseline rather than take the red X at face value:

- **Spellcheck:** the merge *reduces* findings from 44 to 29. It initially added exactly two, both in the reworked Simulations page: `commiting` (a genuine typo, now fixed to "committing") and `WRMSNorm` (a legitimate ODE-solver identifier, now added to `wordlist.txt`). After `c802c20` the diff against the baseline is empty — the merge adds no misspelling that `v13` did not already have.
- **markdown-link-check:** the diff against the baseline is also empty. One earlier run flagged `https://doi.org/10.1201/b17625` as a 404; that URL sits in `references.md`, which this merge does not touch, and is present identically on `v13`. It resolved on the next run — transient external-link flake, not a regression.

The spell-check wordlist itself resolved to v13's, which is a superset of v12's, so #312's prose stays covered.

## One thing to look at that did *not* conflict

`SUMMARY.md` auto-merged to v12's navigation order, which moves **Parameters, Formulas, Tags, and Keywords** from just after *Initial Conditions* down to just before *Simulations*. The accompanying rename *Setting up a Simulation* → *Simulations* follows #312 and is correct; the reordering is a silent side effect, left as-is rather than overridden in either direction, but worth a conscious decision.


Yuri05:
@PavelBal Is this OK?

image

PavelBal:
@Yuri05 yes, it's even better so.

Yuri05 (approved):

PR #368, opened 2026-08-13
Document "New Condition Group (AND/OR)" in condition lists

PavelBal:
Closes #353

Expands the section ["How Tags are used"](https://github.com/Open-Systems-Pharmacology/docs/blob/docs/condition-group/part-4/parameters-formulas-tags.md#how-tags-are-used---container-criteria-for-formulas-observers-transports-and-events) in *Parameters, formulas, tags, and keywords* to document condition groups:

- Clarifies that a condition list is combined with **one** top-level operator (AND or OR), switchable via the context menu.
- Adds a new item **7. Condition group** describing:
  - the purpose: mixing AND and OR logic within one criterion, e.g. `(Tagged with "Venous Blood" AND tagged with "Plasma") OR (tagged with "Muscle" AND tagged with "Interstitial")`;
  - how to create one (right-click → **New condition group (AND/OR)** → **Create condition group** dialog);
  - how the group is merged with the rest of the list: the group is a single parenthesized entry — its inner conditions are combined by the group's operator, and the result is combined with the sibling conditions by the list's operator;
  - a worked example based on the section's existing `Organism / Container A / Container B` example model;
  - the limitation that condition groups cannot be nested.
- Updates the closing paragraph accordingly.

Behavior was verified against the implementation (`ConditionGroup` in OSPSuite.Core, `CreateConditionGroupPresenter`/`TagTask` in MoBi).

No screenshot of the "Create condition group" dialog is included — one can be added in a follow-up if desired.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Documentation**
  * Clarified how tag conditions use AND or OR logic.
  * Added guidance for combining mixed AND/OR logic through condition groups.
  * Documented condition-group creation, evaluation order, examples, and one-level nesting limitations.

Issue #356, opened 2026-08-11
Observers: container criteria merge behavior under "Extend" is documented incorrectly


PavelBal:
## Description

[part-4/modularization-concept.md](https://github.com/Open-Systems-Pharmacology/docs/blob/master/part-4/modularization-concept.md), section **Observers → Merge behavior "Extend"**, states:

> - The **Operator** of the "In container with" list is overwritten.
> - The **Conditions** list of the "In container with" list is extended.

This does not match the current implementation. In OSPSuite.Core (`origin/develop` / `V13`), `SimulationBuilder.mergeObservers` only merges the molecule lists and the formula — it does **not touch** the observer's `ContainerCriteria` at all. `ObserverBuilder` is not a container, so the generic container merge does not apply to it either.

As a result, when a later module extends an equally-named observer, the **earlier (base) module's container criteria are kept unchanged** — the extending module's criteria (both conditions and operator) are silently ignored.

## Expected documentation

Under "Extend", the observer's container criteria should be documented as **not merged**: the criteria of the module higher in the hierarchy win, and criteria changes from an extending module have no effect.

## References

- Code: `SimulationBuilder.cs` (`mergeObservers`) on OSPSuite.Core `develop`
- Related: the Events section was already corrected for the analogous behavior change (#2811 in OSPSuite.Core)

Issue #355, opened 2026-08-11
Reactions: container criteria merge behavior under "Extend" is documented incorrectly


PavelBal:
## Description

[part-4/modularization-concept.md](https://github.com/Open-Systems-Pharmacology/docs/blob/master/part-4/modularization-concept.md), section **Reactions → Merge behavior "Extend"**, states:

> **Container criteria** are extended, the **Operator** is overwritten.

This does not match the current implementation. In OSPSuite.Core (`origin/develop` / `V13`), `SimulationBuilder.mergeReactions` **fully overwrites** the reaction's `ContainerCriteria` with the later module's criteria:

```csharp
if (incoming.ContainerCriteria != null)
   targetReaction.ContainerCriteria = incoming.ContainerCriteria;
```

Since `ReactionBuilder` initializes `ContainerCriteria` to `new DescriptorCriteria()`, it is never `null`, so the criteria are **always completely overwritten** by the later module (an empty incoming criteria means "everywhere"). Conditions are not combined and the operator is not treated separately.

## Expected documentation

Under "Extend", the reaction's container criteria should be documented as **overwritten** by the module lower in the hierarchy, not extended.

## References

- Code: `SimulationBuilder.cs` (`mergeReactions`) on OSPSuite.Core `develop`
- Related: the Events section was already corrected for the analogous behavior change (#2811 in OSPSuite.Core)

PR #369, opened 2026-08-13
Document how condition lists are combined when merging modules

### LINKED ISSUE #357: How are container criteria and other logical lists extended/combined?
Container criteria etc. If there is only one group, the operator is overwritten in both modes. What happens when there are multiple groups?

PavelBal:
Closes #357.

Adds a **Condition lists** section to *Creating simulations from modules and combination rules*, describing what "extended" means for the container criteria of observers, passive transports, events, and reactions — including condition groups, which were documented in #368 but whose behavior when modules are combined was not covered.

The rules were derived from `SimulationBuilder.mergeDescriptorCriteria` in OSPSuite.Core:

```csharp
target.Operator = source.Operator;
source.Each(t => target.Add(t.CloneCondition()));
```

- Conditions of the later module are appended; nothing is ever removed, and duplicates are not filtered out (`DescriptorCriteria` is a plain `List<ITagCondition>`).
- A condition group is an `ITagCondition` in its own right (`ConditionGroup : DescriptorCriteria, ITagCondition`), so it is appended as a single entry and `CloneCondition()` deep-clones it with its own operator and conditions. Groups of two modules are therefore never combined with each other.
- The operator of the list itself is overwritten unconditionally, regardless of how many conditions either module contributes.

The warning covers the practical trap behind the issue: a list holding a single condition has no visible operator in MoBi but still stores one, and it defaults to `And` (`DescriptorCriteria.Operator`). An extension module that adds one condition can therefore silently turn an earlier module's `OR` list into an `AND` list, which changes the set of containers the entity is created in.

Scope notes:

- The section is deliberately worded as "wherever the sections below state that such a list is *extended*", so it stays correct independently of #355 and #356, where the per-building-block statements for reactions and observers are being corrected. Those are tracked as implementation bugs in Open-Systems-Pharmacology/MoBi#2490 and Open-Systems-Pharmacology/MoBi#2491, and only the sections concerned will need an update once they are resolved.
- `ConditionGroup` was added to OSPSuite.Core after the module merge logic (Open-Systems-Pharmacology/OSPSuite.Core#2845), and no test covers a condition group going through a module merge. What is documented here is the behavior of the current implementation — if appending groups is not the intended semantics, this section should be revisited together with the code.

## Summary by CodeRabbit

* **Documentation**
  * Added guidance on how condition lists are merged or overwritten across modules.
  * Documented how operators are selected during merges.
  * Added a warning that operator changes may alter the model structure.

### DIFF
--- a/part-4/modularization-concept.md
+++ b/part-4/modularization-concept.md
@@ -66,6 +66,22 @@ During simulation creation, the modules are combined to a common model structure
 
 There are two types of combination behavior that can be defined for a module - **overwrite** or **extend**. The following sections describe how different building blocks are merged and what are the differences between the **overwrite** and the  **extend** modes, if any.
 
+### Condition lists
+
+Observers, passive transports, events, and reactions use *condition lists* - the lists titled "In container with" or "Between containers with" - to define where in the model structure they are created (see [How Tags are used](parameters-formulas-tags.md#how-tags-are-used---container-criteria-for-formulas-observers-transports-and-events)). Wherever the sections below state that such a list is *extended*, the following rules apply:
+
+- All conditions defined in module `B` are added to the conditions defined in module `A`. Conditions are never removed. A condition defined in both modules is added twice, which does not change the result of the evaluation.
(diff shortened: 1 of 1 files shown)


PavelBal:
@coderabbit review

Yuri05 (approved):

PR #371, opened 2026-08-13
Fix the order in which parameter values are applied

PavelBal:
Item 5 of the re-review in #335.

The **Parameter values** section lists the individual as step 2 and the expression profile as step 3. Since the list is introduced as "determined in the following order" — and steps 3 and 4 are worded as later-wins ("overriding the value from the Molecules BB", "the value from the latest module is selected") — it reads as *expression profile beats individual*. The implementation is the other way round (`OSPSuite.Core/Domain/Services/QuantityValuesUpdater.cs:57-72`):

```csharp
updateMoleculeAmountFromInitialConditions(modelConfiguration);

//Add expressions profile before individual as some settings might be overwritten in the individual for aging
updateParameterFromExpressionProfiles(valueUpdater);

updateParameterFromIndividualValues(valueUpdater);

//PV are applied last
updateParameterValueFromParameterValues(valueUpdater);
```

So for a parameter defined in both, **the value from the individual wins**, deliberately.

This is not a theoretical difference. PK-Sim normally keeps expression and ontogeny parameters out of the individual building block (`IndividualToIndividualBuildingBlockMapper.shouldExportParameter` filters them), but the aging conversion writes them back in: `DistributedParameterToTableParameterConverter.createParameterValueVersionOntogenyTableParameter` puts `<Protein>|Ontogeny factor` / `Ontogeny factor GI` into the individual as a `TableFormula` with `Value = null`, while the expression profile carries the same paths as constants. If the documented order were the implemented one, every aging simulation would lose the ontogeny time-dependence.

This PR:

- swaps steps 2 and 3 so the sequence matches the implementation (Building Block → Expression Profile → Individual → PV BB), keeping the text of each step;
- makes the semantics of the list explicit — "a value applied later overwrites a value applied before it" — which is what made the inversion easy to miss;
- adds one paragraph explaining why the individual has to win, so the ordering does not look like an error to the next reader;
- corrects the same sequence in the "Unchanged" summary of the v12→v13 page.

The order itself has been in place since the modularization work in 2023 (`3c18ced2`), so the migration page's statement that the resolution order is *unchanged* between v12 and v13 remains correct — only the sequence it quotes was wrong.

No overlap with #359 or #370.

### DIFF
--- a/part-4/converting-v12-projects-to-v13.md
+++ b/part-4/converting-v12-projects-to-v13.md
@@ -164,7 +164,7 @@ This first-module precedence is considered a bug — the administered molecule s
 
 ### Unchanged
 
-- **Parameter Values** — resolution order (Building Block → Individual → Expression Profile → PV BB, latest module wins) is unchanged.
+- **Parameter Values** — resolution order (Building Block → Expression Profile → Individual → PV BB, latest module wins) is unchanged.
 - **Initial Conditions** — resolution order (Molecules BB → Expression Profiles → IC BBs, latest module wins) is unchanged.
 
 ## Migration steps
--- a/part-4/modularization-concept.md
+++ b/part-4/modularization-concept.md
@@ -252,15 +252,17 @@ For each event:
 
 ### Parameter values
 
-The final values or formulas of the parameters in a simulation are determined in the following order:
+The final values or formulas of the parameters in a simulation are determined in the following order. The values are applied in this sequence, so a value applied later overwrites a value applied before it:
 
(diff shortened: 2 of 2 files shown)


part-4/converting-v12-projects-to-v13.md
```diff
@@ -164,7 +164,7 @@ This first-module precedence is considered a bug — the administered molecule s
 
 ### Unchanged
 
-- **Parameter Values** — resolution order (Building Block → Individual → Expression Profile → PV BB, latest module wins) is unchanged.
+- **Parameter Values** — resolution order (Building Block → Expression Profile → Individual → PV BB, latest module wins) is unchanged.
```
Yuri05, line 167: assuming this is correct description :)

Yuri05 (approved):

PR #381, opened 2026-08-21
Replace relative documentation links with absolute ones in NEWS.md

Yuri05:
Thanks for asking me to work on this. I will get started on it and keep this PR's description up to date as I form a plan and make progress.

> ----
> 
> *This section details on the original issue you should resolve*
> 
> <issue_title>Replace relative documentation links with absolute ones in NEWS.md</issue_title>
> <issue_description>In https://github.com/Open-Systems-Pharmacology/docs/blob/v13/NEWS.md : 
> replace all relative documentation links with absolute links to the published v13 documentation. 
> 
> E.g. replace 
> `[Converting v12 MoBi projects to v13](part-4/converting-v12-projects-to-v13.md)` 
> with 
> `[Converting v12 MoBi projects to v13](https://docs.open-systems-pharmacology.org/v13/working-with-mobi/mobi-documentation/reuse-of-project-information-from-previous-versions/converting-v12-projects-to-v13)`
> 
> This would allow copy/paste of the markdown source into release notes etc.</issue_description>
> - Closes Open-Systems-Pharmacology/docs#378
> 
> ## Comments on the Issue (you are @claude[agent] in this section)
> 
> <comments>
> </comments>
- Fixes #378

PR #394, opened 2026-09-02
Document selecting the module and building block when committing simulation changes

PavelBal:
Documents [MoBi#1614](https://github.com/Open-Systems-Pharmacology/MoBi/issues/1614) (implemented in [MoBi#2526](https://github.com/Open-Systems-Pharmacology/MoBi/pull/2526)) and adds the NEWS entry that #388 was missing.

## Commit target selection (MoBi#1614)

Simulation changes were always committed to the building blocks of the **last** module in the configuration. MoBi now asks for the target instead: the module, and within it the Parameter Values / Initial Conditions building block - or a newly created one.

Both places that stated the old rule are updated:

- [`part-4/modularization-concept.md`](https://github.com/Open-Systems-Pharmacology/docs/blob/docs/v13-commit-target-selection/part-4/modularization-concept.md#committing-changes-from-a-simulation-to-a-building-block) - full description with the dialog screenshot, the fields it offers, and the two cases in which MoBi warns that the simulation stays out of sync: a target building block the simulation does not use, and committed values shadowed by a later module.
- [`part-4/setting-up-simulation.md`](https://github.com/Open-Systems-Pharmacology/docs/blob/docs/v13-commit-target-selection/part-4/setting-up-simulation.md#commit-to-building-blocks) - shortened description with a link to the section above.

The warning hint about PK-Sim modules being converted to extension modules now says "the **selected** PK-Sim module" instead of "the last PK-Sim module".

## NEWS

- New subsection **MoBi®: Selecting where simulation changes are committed**.
- Bullet for the **Warn for non-finite quantities** option documented in #388 - that PR only touched the Options page, so NEWS had no entry for it.

Both use absolute documentation links, following #381.

## Please confirm

The dialog sequence. MoBi#2526 describes "a single dialog", while the screenshot shows *Select Module and Building Blocks* on its own after the change-selection dialog. Both pages currently word it as a second dialog that opens after selecting which changes to commit - please correct if the two are actually merged into one.

## Summary by CodeRabbit

- **New Features**
  - Simulation changes can now be committed to a selected module and parameter values or initial conditions building block.
  - Added warnings for commits that leave the simulation out of sync.
  - Added an option to report non-finite quantities, such as `NaN` or infinity, at the start of a simulation.

- **Documentation**
  - Updated guidance on committing simulation changes and clarified related warnings and selection behavior.


PavelBal:
@coderabbitai review

part-4/modularization-concept.md
```diff
@@ -300,13 +300,23 @@ It is no longer possible to update only selected building blocks.
...
+After selecting which changes to commit, a dialog opens where the target of the commit is selected:
+
+![Selecting the module and the building blocks that receive the committed changes](../assets/images/part-4/MoBi-SelectCommitTarget.png)
+
+- **Module** - the module that receives the changes. One module is selected for all changed building block types, so committed changes always stay together in the same module.
+- **Parameter values building block** and **Initial conditions building block** - the Parameter Values (PV) or Initial Conditions (IC) building block of the selected module that receives the changes, or "New Parameter Values Building Block"/"New Initial Conditions Building Block" to create one. A selection is only offered for the building block types that actually have changes, and the lists are updated when the module selection changes.
+
+By default, the dialog offers the last module of the simulation configuration and the building blocks the simulation currently uses. Selecting **Cancel** aborts the commit. If there is nothing to choose - a single module with a single possible target for every changed building block type - the dialog is not shown.
```
rwmcintosh: I think the no-choice is only offered if there is a single module and the module does not contain any PV (or IC). Because with a single, selected or unselected BB, the user could select to create a new BB, or use the existing one.

In other words, the existence of a matching type BB means there is a choice to make.

part-4/modularization-concept.md
```diff
@@ -300,13 +300,23 @@ It is no longer possible to update only selected building blocks.
 
 Changes to parameter or start values made in a simulation can be committed to the building block (BB) the parameter/start value belongs to. To do so, right-click on the simulation and select "Commit to building blocks".
```
Yuri05: should we replace **start value(s)** with **(molecule) initial value(s)** in this sentence?

part-4/modularization-concept.md
```diff
@@ -300,13 +300,23 @@ It is no longer possible to update only selected building blocks.
...
+
+By default, the dialog offers the last module of the simulation configuration and the building blocks the simulation currently uses. Selecting **Cancel** aborts the commit. If there is nothing to choose - a single module with a single possible target for every changed building block type - the dialog is not shown.
 
 {% hint style="warning" %}
-When committing changes made to a simulation created from PK-Sim module(s) only, the last PK-Sim module will be converted to an extension module. To avoid this, create a new extension module that should contain the changes. Then create the simulation with the extension module.
+When committing changes made to a simulation created from PK-Sim module(s) only, the selected PK-Sim module will be converted to an extension module. To avoid this, create a new extension module that should contain the changes. Then create the simulation with the extension module.
```
Yuri05, line 313: Reading this: not for now, but maybe for v13.1: a nice add-on would be the possibility to create a new empty module directly from the "Select modules and BBs" dialog. Then if a simulation was created from the only PK-Sim module, we still could commit parameter changes to a new module without the need to manually create a new module and reconfigure the simulation
@rwmcintosh If it's an easy change, I would consider doing this even right now

part-4/modularization-concept.md
```diff
@@ -300,13 +300,23 @@ It is no longer possible to update only selected building blocks.
 
 Changes to parameter or start values made in a simulation can be committed to the building block (BB) the parameter/start value belongs to. To do so, right-click on the simulation and select "Commit to building blocks".
 
-Changes to parameters values are committed to the selected Parameter Values (PV) building block of the last module in the simulation configuration. If the last module does not contain a PV BB, or no PV BB has been selected for this module, new PV BB will be created.
+After selecting which changes to commit, a dialog opens where the target of the commit is selected:
+
```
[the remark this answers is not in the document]
PavelBal: There are no two dialogues, only one. Updated sentence.

part-4/modularization-concept.md
PavelBal: Yeah that was wrong

part-4/modularization-concept.md
PavelBal: Changed to "Start values and initial conditions"

Yuri05 (approved):

PR #395, opened 2026-09-02
Document the "Overwrite Parameter Sets" feature

### LINKED ISSUE #360: Document the "Overwrite Parameter Sets" feature

rwmcintosh:
Fixes #360

Documents the Compound Overwrite Parameter Sets introduced in v13 ([PK-Sim #3422](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/3422)).

## New page

`part-3/pk-sim-overwrite-parameter-sets.md`, listed in `SUMMARY.md` after the two Compounds pages:

- Motivation, and how a set differs from **Commit to Building Block**
- Which simulation parameters qualify for a commit
- The orange uncommitted-changes indicator on the simulation and on a compound
- The commit workflow and the dialog, including what happens when an existing set is updated
- The **Overwrite Parameter Sets** tab of a compound: master-detail layout, editing, deleting, default flag, metadata
- Selecting a set for a simulation, and the error on unresolved paths
- Population simulations
- Compound renaming, templates and snapshots, update/commit of the compound, comparison, MoBi export

## Cross-references

- Compounds page: new **Overwrite Parameter Sets tab** section after **Advanced Parameters**, and the tab list updated
- Simulations page: the selection drop-down under **Review compound settings**, and the commit action under **Synchronization options for building blocks in a simulation**

## Release notes

Adds the four deferred screenshots from #375 and links the entry to the new page.

The commit-dialog bullet described the dialog as showing *"parameters grouped by compound with checkboxes"*, which follows the original PRD rather than what shipped. The action is invoked from the context menu of a single compound in a simulation and commits that compound only, so the wording is corrected. The remaining `TODO (screenshot)` markers in `NEWS.md` belong to the PBBM and events-in-protocols entries and are out of scope here.

## Notes for review

- All statements were verified against the PK-Sim `develop` sources rather than the PRD, since the two differ in places.
- The compound editor tab is captioned **ADME** in the UI, while the manual has long said **ADME Properties**, including a section heading. Left unchanged here to avoid an inconsistency within the page; worth a separate fix.

## Summary by CodeRabbit

* **Documentation**
  * Added comprehensive documentation for Overwrite Parameter Sets in PK-Sim®.
  * Documented saving simulation-calculated compound parameters as named sets for reuse.
  * Documented applying parameter sets to individual and population simulations.
  * Added guidance on editing, updating, synchronization, comparison, templates, snapshots, and MoBi® export.
  * Updated compound and simulation workflow documentation with the new tab, actions, and configuration options.
  * Replaced screenshot placeholders with embedded images and added relevant documentation links.


rwmcintosh:
@coderabbitai review please

part-3/pk-sim-overwrite-parameter-sets.md
```diff
@@ -0,0 +1,153 @@
...
+Compounds that are created inside the simulation but are not part of the simulation configuration — for example an unnamed metabolite of a parent/metabolite simulation — are therefore not considered.
+
+Parameters that were applied from an Overwrite Parameter Set are also offered for a commit if you change them again, so a set can be refined step by step.
+
+## Uncommitted changes
+
+As soon as you change such a parameter, PK-Sim® marks the change as **uncommitted**. This is shown by an orange indicator:
+
+- On the **simulation** in the **Simulation Explorer**, an orange overlay is added to the simulation icon as long as the simulation contains any uncommitted compound-dependent change.
+- On the **compound inside the simulation**, the status icon becomes orange in addition to its usual green or red synchronization state. Green/orange means that the compound itself is in sync with its building block but has uncommitted compound-dependent simulation parameter changes; red/orange means that it is out of sync *and* has uncommitted changes.
```
Yuri05: ..._the status icon becomes orange in addition to its usual green or red synchronization state_....
This sounds for me as an **existing** red/green icon would become orange, which is not the case.
Would reformulate ("..._a second orange status indicator is added in addition to the green or red synchronization state_..." or so)
And then maybe "_...combination of green and orange status indicators means..._" instead of "_...green/orange means..._" (same for red/orange)

Yuri05: The screenshot below makes it clear, but the text is confusing

part-3/pk-sim-overwrite-parameter-sets.md
```diff
@@ -0,0 +1,153 @@
...
+
+![The context menu of a compound used in a simulation, with the Commit Simulation Parameters to Compound action.](../assets/images/part-3/overwrite-parameter-sets-context-menu.png)
+
+{% hint style="info" %}
+The action is only offered for a compound of an **individual simulation** that actually has uncommitted changes. Overwrite Parameter Sets can be *used* in population simulations, but they cannot be created from one — see [Population simulations](#population-simulations) below.
+{% endhint %}
+
+The **Commit simulation parameters to compound** dialog opens. It contains:
+
+- A grid of all uncommitted parameters of this compound, with the columns **Selected**, **Parameter** (the path of the parameter in the simulation), **Value** (with its display unit) and **Value Origin**. Every parameter is selected by default; clear the check box of the ones you do not want to store.
```
Yuri05: "A table" instead of "A grid"?

part-3/pk-sim-overwrite-parameter-sets.md
```diff
@@ -0,0 +1,153 @@
...
+
+- Parameters you are committing are added to the set, or their stored value is overwritten.
+- Parameters that were in the set and that you have **reset** in the simulation since the last commit are **removed** from the set.
+- Parameters that were in the set and that you have not touched are **preserved**.
+
+After an update commit, re-creating the simulation with this set therefore reproduces exactly the parameter values the simulation has at the moment of the commit.
+
+## The Overwrite Parameter Sets tab of a compound
+
+The Overwrite Parameter Sets of a compound are shown in the **Overwrite Parameter Sets** tab of the compound editor, after the **Advanced Parameters** tab. The tab uses a master-detail layout:
```
Yuri05: "_The Overwrite Parameter Sets of a compound are shown in the Overwrite Parameter Sets tab of the compound editor_"
Tautologie... would cheange to smthg like
_The Overwrite Parameter Sets of a compound are shown in the same-named tab of the compound editor_.

part-3/pk-sim-overwrite-parameter-sets.md
```diff
@@ -0,0 +1,153 @@
...
+
+- The **list on the left** shows all Overwrite Parameter Sets of the compound in the columns **Name**, **Default**, **Species** and **Disease State**, followed by a button that deletes the set.
```
Yuri05: ...**in** the columns... ==> ...**with** the columns... ?

part-3/pk-sim-overwrite-parameter-sets.md
```diff
@@ -0,0 +1,153 @@
...
+- The **details on the right** show the **Metadata** of the selected set above a grid of its parameter values, in the columns **Path**, **Value** and **Value Origin**, followed by a button that deletes the entry. The value is displayed together with its unit.
```
Yuri05: _grid_=>_table_

part-3/pk-sim-overwrite-parameter-sets.md
```diff
@@ -0,0 +1,153 @@
...
+
+- Change the **Value** of a stored parameter.
+- Change the unit of the value, in the unit selector of the value editor. As for any other parameter in the suite, the displayed number is kept and the value in base unit is recalculated (see [Default, Display and Base Units](../part-5/default-display-base-units.md)).
+- Set the **Value Origin** of the entry, to document where the value comes from.
+- Delete a single parameter from the set, with the button at the end of its row.
+
+Additionally you can:
+
+- Delete an entire Overwrite Parameter Set, with the button at the end of its row in the list on the left, after confirming the deletion.
+- Mark a set as the **Default** of the compound, or clear that flag. A compound can have at most one default set, and setting a new default clears the previous one. A compound is not required to have a default set.
```
Yuri05: Would add that the default set is preselected when creating a new simulation. No default set - none is preselected.

part-3/pk-sim-overwrite-parameter-sets.md
```diff
@@ -0,0 +1,153 @@
...
+
+All of these changes are regular PK-Sim® commands and can be undone.
```
Yuri05: would remove "_are regular PK-Sim® commands_" (from the users perspective it's unclear what is meant here)

part-3/pk-sim-overwrite-parameter-sets.md
Yuri05: OK, it's described in the next sub-chapter. Still I think I would mention it here with the reference to the "Create Simulation" subchapter.

part-3/pk-sim-overwrite-parameter-sets.md
```diff
@@ -0,0 +1,153 @@
...
+Midazolam|Intestinal permeability (paracellular)
+```
+
+The set can therefore only be applied to a simulation in which exactly these paths exist.
+
+Parameters that were overwritten this way behave as **compound parameters** from then on:
+
+- They are shown as compound parameters and lose the reset behavior of a simulation parameter, because their reference value is now the value stored in the set.
+- Their value origin is taken from the entry of the set.
+- They have no counterpart in the Compound building block itself, so they can only be written back through a commit to an Overwrite Parameter Set, not through **Commit to Building Block**.
```
Yuri05, line 146: This is wrong, right? 
Even in this docu below it states:
"Update and commit of the compound. Update from Building Block and Commit to Building Block on a compound of a simulation also synchronize the Overwrite Parameter Sets between the compound in the simulation and the Compound building block of the project."

NEWS.md
```diff
@@ -178,26 +178,28 @@ Events - e.g., meal intake, gallbladder emptying - can now be defined directly w
...
 
-- **Committing simulation parameters to the compound** - a new "Commit Simulation Parameters to Compound" action is available for each compound used in a simulation. This action opens a dialog listing all changed parameters, which can be deselected here. Then, either a new "Overwrite Parameter Set" can be created with its own name, or an existing one can be updated.
+- **Committing simulation parameters to the compound** - a new "Commit Simulation Parameters to Compound" action is available in the context menu of each compound used in a simulation. This action opens a dialog listing the changed parameters of that compound, which can be deselected here. Then, either a new "Overwrite Parameter Set" can be created with its own name, or an existing one can be updated. Each compound is committed separately, from its own node in the simulation.
 
-> 🖼️ **TODO:** Screenshot of the "Commit simulation parameters to compounds" dialog, showing parameters grouped by compound with checkboxes and name/dropdown options.
+![The Commit simulation parameters to compound dialog.](assets/images/part-3/overwrite-parameter-sets-commit-dialog.png)
```
Yuri05: In **NEWS.md**, please use _fully qualified links_: "https://docs.open-systems-pharmacology.org/v13/..."
This would allow to copy/paste the content of NEWS.md to release notes/forum/etc. AS IS (otherwise all those links would become broken outside of the documentation page).

Yuri05 (changes requested):
s. comments

part-3/pk-sim-overwrite-parameter-sets.md
rwmcintosh: I think that part has recently been removed... You have to update manually to get the updated OPS into the simulation compound building block.

So, yeah, one is wrong. let me do the steps again.

part-3/pk-sim-overwrite-parameter-sets.md
rwmcintosh: Reworded:

> On the **compound inside the simulation**, a second, orange status indicator is added next to the green or red synchronization state, which itself does not change. The combination of a green and an orange indicator means that the compound is in sync with its building block but has uncommitted compound-dependent parameter changes; the combination of a red and an orange indicator means that it is out of sync *and* has uncommitted changes.

part-3/pk-sim-overwrite-parameter-sets.md
rwmcintosh: Done.

part-3/pk-sim-overwrite-parameter-sets.md
rwmcintosh: Reworded:

> The Overwrite Parameter Sets of a compound are shown in the same-named tab of the compound editor, after the **Advanced Parameters** tab.

part-3/pk-sim-overwrite-parameter-sets.md
rwmcintosh: Done.

part-3/pk-sim-overwrite-parameter-sets.md
rwmcintosh: Done.

part-3/pk-sim-overwrite-parameter-sets.md
rwmcintosh: Added, with the reference:

> Mark a set as the **Default** of the compound, or clear that flag. A compound can have at most one default set, and setting a new default clears the previous one. A compound is not required to have a default set. The default set is the one preselected when the compound is used in a new simulation; if the compound has no default set, no set is preselected (see [Using an Overwrite Parameter Set in a simulation](#using-an-overwrite-parameter-set-in-a-simulation)).

part-3/pk-sim-overwrite-parameter-sets.md
rwmcintosh: Removed:

> All of these changes can be undone.

NEWS.md
rwmcintosh: Done - the four image links and the link to the documentation page are now fully qualified. Images point at `https://raw.githubusercontent.com/Open-Systems-Pharmacology/docs/v13/assets/images/part-3/...`, since `docs.open-systems-pharmacology.org` does not serve the raw files; happy to switch if you prefer a different host.

Note that this makes `markdown-link-check` fail on this PR with 5 dead links - the four images and the new page - because none of those targets exist on `v13` until this is merged. They resolve after the merge. All other checks pass.

part-3/pk-sim-overwrite-parameter-sets.md
rwmcintosh: Line 132 is correct — the bullet below it was the wrong one. The behaviour is the one from https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/3684: a commit writes the set to the project compound only.

**Before**

> **Update and commit of the compound.** **Update from Building Block** and **Commit to Building Block** on a compound of a simulation also synchronize the Overwrite Parameter Sets between the compound in the simulation and the Compound building block of the project.

**After**

> **Update and commit of the compound.** A commit writes the Overwrite Parameter Set to the Compound building block of the project, not to the copy of the compound held by the simulation. Both **Update from Building Block** and **Commit to Building Block** copy the Overwrite Parameter Sets of the building block into the compound of the simulation, where they can then be selected. Sets are never copied in the opposite direction.

NEWS.md
Yuri05: I think it's OK like this

part-3/pk-sim-overwrite-parameter-sets.md
Yuri05: Still confusing for me.
> "_A commit writes the Overwrite Parameter Set to the Compound building block of the project, not to the copy of the compound held by the simulation._" 

⬆️ This is what a normal commit (for compound or for any other BB type) does as well.

> "_Both Update from Building Block and Commit to Building Block copy the Overwrite Parameter Sets of the building block into the compound of the simulation, where they can then be selected. Sets are never copied in the opposite direction._" 

Uff. Totally confusing, sorry. Maybe we should go here use case by use case and explain what happens.
a) Committing of the compound to BB if the changes were made to an OPS parameter in compound
b) Committing of the compound to BB after simulation parameters were committed to an OPS which is NOT currently used in the simulation
c) Committing of the compound to BB after simulation parameters were committed to an OPS which is used by the simulation (2 sub-cases: committing to the OPS used by the simulation and committing to another OPS (new or existing)
d) Updating from the compound BB if an OPS parameter was changed **in the OPS in compound**
e) Updating from the compound BB if an OPS parameter was changed **in the simulation**
f)... (more?)
Once we have the clear description of the all the use cases, we can think if we should keep that description AS IS or compress somehow to a "shorter" one

part-3/pk-sim-compounds-definition-and-work-flow.md
```diff
@@ -427,25 +427,9 @@ Four drug-related parameters which are used in the model for proteins and large
 
 Some parameters that depend on the compound are not part of the compound building block because they are only created in the context of a simulation, e.g., permeabilities or partition coefficients of the organs. If such parameters have been modified in a simulation, they can be committed back to the compound as a named **Overwrite Parameter Set** and re-applied in other simulations and projects.
 
-**Committing simulation parameters to a compound**
+The **Overwrite Parameter Sets** tab lists all sets defined for the compound in a master-detail view. Parameter values, units, and metadata (e.g., species or disease state) can be inspected and edited, individual parameters or whole sets can be deleted, and one set can be marked as **Default**. New sets are only created by committing from a simulation, never from this tab.
```
PavelBal: "Master-detail view" is not a common term, I don't know what it means.

part-3/pk-sim-overwrite-parameter-sets.md
rwmcintosh: Went through the cases. They all follow from two rules, which I verified with integration specs on a real simulation:

**R1** — Both *Update from Building Block* and *Commit to Building Block* make the compound in the simulation an exact mirror of the overwrite parameter sets on the Compound building block: missing sets are added, differing ones updated, ones no longer on the building block removed. The direction is always building block → simulation, for both operations (covered by the specs from #3685 / #3686).

**R2** — Neither operation touches a simulation parameter value that came from a set. Values from a set are applied only when the simulation is created or configured.

| case | set on the compound | set in the compound of the simulation | simulation parameter value |
|---|---|---|---|
| a) Commit to BB, value changed in the set in the compound | keeps the changed value, it is *not* overwritten from the simulation | mirrored to the changed value | unchanged |
| b) Commit to BB after committing simulation parameters to a set **not** used by the simulation | contains the new/updated set | mirrored, the set becomes selectable | unchanged, the set is not selected |
| c1) Commit to BB after committing simulation parameters to the set **used** by the simulation | holds the values the simulation had at the commit | mirrored | unchanged — it already holds those values |
| c2) Commit to BB after committing simulation parameters to **another** set (new or existing) | that other set holds the values, the selected one is untouched | mirrored | unchanged, but see the note |
| d) Update from BB, value changed in the set in the compound | changed value | mirrored to the changed value | **unchanged** — the new value only takes effect when the simulation is created or configured again |
| e) Update from BB, an OPS parameter changed **in the simulation** | unchanged | unchanged | **the change is kept** and stays an uncommitted change |

Measured: (a) set 183.5→192.5, after commit template 192.5 / simulation copy 192.5 / parameter 183.5. (d) set 183.5→190.5, after update simulation copy 190.5 / parameter 183.5. (e) parameter 183.5→188.5, after update parameter 188.5, still tracked.

Note on c2: committing untracks the committed paths, so the orange indicator disappears even though the simulation's values no longer match its own selected set. Re-creating or configuring the simulation would revert them to the selected set. This may deserve an explicit warning in the documentation.

For f) I would add two that are already in the page: a set cannot be deleted while any simulation uses it, and renaming the compound also renames the compound in the paths stored in all its sets.

Since every case reduces to R1 + R2, my suggestion is to state those two rules in the page rather than the full table.

Yuri05: I strongly vote for keeping the full table.
I mean, you could bring the two rules first, and then illustrate with the table what in means in a concrete situation.

Yuri05: But maybe @PavelBal could comment first to say whether this explanation answers all his recent questions without creating new ones. 😉

rwmcintosh:
PK-Sim issues opened out of this review:

- Open-Systems-Pharmacology/PK-Sim#3733 - configuring a simulation clears the uncommitted compound parameter changes, so they can no longer be committed
- Open-Systems-Pharmacology/PK-Sim#3734 - commit dialog: "Update Existing Parameter Set" disabled and a duplicate set name accepted
- Open-Systems-Pharmacology/PK-Sim#3735 - track resets of overwrite parameter set values so that they can be committed

I suggest these changes for V13. The documentation would follow once they are implemented.

Open-Systems-Pharmacology/PK-Sim#3731 has been closed in favour of this pull request, so that the discussion stays in one place.

part-3/pk-sim-overwrite-parameter-sets.md
rwmcintosh: Keeping the table. The page now has a dedicated section *Updating and committing the compound* with the two rules followed by the case table.

Yuri05 (approved):

PR #396, opened 2026-09-03
Document MoBi snapshots

### LINKED ISSUE #365: Document MoBi snapshots

rwmcintosh:
Fixes #365

Documents MoBi snapshots, and promotes the PK-Sim snapshot section to a page of its own since snapshots are now offered for every building block type.

## New pages

- **`part-4/mobi-snapshots.md`** — the MoBi project snapshot introduced in v13: what it stores (PK-Sim modules as embedded PK-Sim snapshots re-created by a local PK-Sim installation, extension modules as PKML, simulations as their configuration plus user-changed values), what it does not store, export/load workflows, and the per-object **Snapshot** menu on Modules, Individuals and Expression Profiles that originate from PK-Sim.
- **`part-3/pk-sim-snapshots.md`** — the former "Exporting Project to Snapshot / Loading Project from Snapshot" section, moved out of *Importing and Exporting Project Data and Models* and extended with a **Snapshots of single building blocks** section.

The original section is replaced by a pointer with its heading kept, so the existing anchor still resolves. The six inbound links are repointed at the new page.


part-4/mobi-snapshots.md
```diff
@@ -0,0 +1,99 @@
...
+
+## What a MoBi project snapshot contains
+
+| Project content | How it is stored in the snapshot |
+| --- | --- |
+| PK-Sim® modules | as the PK-Sim® snapshot the module carries, embedded as a JSON sub-node. The module is **re-created by PK-Sim®** when the snapshot is loaded. |
+| Extension modules | as PKML, stored verbatim (Base64-encoded). They are restored as they were, not re-created. |
+| Individuals and Expression Profiles created in PK-Sim® | as the PK-Sim® snapshot of the building block, plus the parameter values (and formulas) that were changed in MoBi® afterwards. The building block is re-created by PK-Sim® and the stored changes are re-applied on top. |
+| Individuals and Expression Profiles not originating from PK-Sim® | as PKML, stored verbatim (Base64-encoded). |
```
PavelBal: Okay for now, but can this be changed in the future? So individuals and expression profiles are also stored as their "snapshots" and can be re-created?

part-4/mobi-snapshots.md
```diff
@@ -0,0 +1,99 @@
...
+The project snapshot also records the MoBi® project version it was written with and the application it was created with.
+
+## What is not stored
+
+The following is not part of a MoBi® project snapshot and is therefore missing after loading:
+
+* **Simulation results.** They are re-computed when the simulations are run (s. below).
+* **Sensitivity Analyses.**
+* **The [Working Journal](../part-5/working-journal.md)** and the [project history](../part-5/history-manager-history-reporting.md).
+* **Chart templates** and the **project-wide default simulation settings** (the default solver settings, output schema and output selections used for new simulations). They are reset to the MoBi® defaults.
```
PavelBal, line 32: Can you open an issue to extend this?

part-4/mobi-snapshots.md
```diff
@@ -0,0 +1,99 @@
...
+
+## Snapshots of single modules and building blocks
+
+Modules, Individuals and Expression Profiles that were created in PK-Sim® carry their own PK-Sim® snapshot. For these, a **Snapshot** entry is available in the context menu of the object in the **Modules Explorer**:
+
+* **Reload module** / **Reload Individual** / **Reload Expression Profile** re-creates the object from its stored PK-Sim® snapshot through the local PK-Sim® installation and **adds it to the project as an additional object** — the existing one is not overwritten. Use it to obtain an unmodified copy, or a copy rebuilt by a newer PK-Sim® version.
+* **Export...** writes the stored PK-Sim® snapshot to a `*.json` file. For a module, this is a complete PK-Sim® project snapshot and can be loaded in PK-Sim®.
+
+{% hint style="info" %}
+The snapshot embedded in a module or building block is the one PK-Sim® wrote when the model was transferred to MoBi®; it is not updated by later changes in MoBi®. Changes made in MoBi® to the parameter values of an Individual or an Expression Profile are therefore **not** contained in the file written by **Export...**, whereas they *are* contained in a snapshot of the whole project.
```
PavelBal, line 85: Also here - new issue, should be supporter later

part-4/mobi-snapshots.md
```diff
@@ -0,0 +1,99 @@
...
+{% endhint %}
+
+{% hint style="info" %}
+The **Snapshot** menu appears only on objects that carry a PK-Sim® snapshot, i.e. on Modules, Individuals and Expression Profiles that originate from PK-Sim®. The remaining MoBi® building block types - Spatial Structures, Molecules, Reactions, Passive Transports, Observers, Events, Initial Conditions and Parameter Values - are not built from PK-Sim® inputs and are exchanged as PKML instead (s. [The Building Block Concept](building-block-concepts.md)). This differs from PK-Sim®, where **every** building block type can be exported to and loaded from a snapshot (s. [PK-Sim Snapshots](../part-3/pk-sim-snapshots.md#snapshots-of-single-building-blocks)).
```
PavelBal: Add a hint that re-creating a PK-Sim snapshot does not automatically re-create the Expression Profiles and the Individual that were created together with the PK-Sim module. The user should do it manually

PavelBal (changes requested):

part-4/mobi-snapshots.md
rwmcintosh: If they're PKSim originals they're stored as snapshot, but in MoBi, there's no snapshot, just PKML.

part-4/mobi-snapshots.md
rwmcintosh: I don't get it. How could it be supported? The exported snapshot is for PK-Sim, but if you make a change to a module for example, the project will export the module as PKML now since it's no longer a PK-Sim module.

part-4/mobi-snapshots.md
rwmcintosh: Opened Open-Systems-Pharmacology/MoBi#2534 — store chart templates and the project-wide default simulation settings in the project snapshot. The **What is not stored** entry here records the current limitation and should be updated when that lands.

part-4/mobi-snapshots.md
rwmcintosh: Opened Open-Systems-Pharmacology/MoBi#2535.

To close the loop on my earlier question — I was reasoning about modules, where you're right that there is nothing to keep in sync: editing a PK-Sim module sets `IsPKSimModule = false` and it is then stored as PKML. But Individuals and Expression Profiles keep their snapshot when edited (`HasSnapshot => !string.IsNullOrEmpty(Snapshot)`, and nothing resets it), and the *project* snapshot already stores the PK-Sim snapshot plus the MoBi-side edits as `UpdatedValues`. The single-object **Export...** just writes `buildingBlock.Snapshot` verbatim, so it drops them. The issue proposes making the single-object export use the same representation the project snapshot already uses. Scoped to those two types, modules excluded.

PavelBal: Okay what I wanted probably cannot be done for Expression Profiles. But at least for individuals? Instead of exporting the individuals as encoded PKML, store its input metadata from which the individual can be recreated.

part-4/mobi-snapshots.md
rwmcintosh: We discovered that I was misunderstanding the situation. Pavel was describing [this bug](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/3741) (which will be fixed in V13).

We eventually came to the conclusion that it's not appropriate to describe lines 15 and 16 as "Individuals and Expression Profiles created in PK-Sim®" All are created in PK-Sim. But to a user, the original exchange looks like it's created in MoBi. So what we want there is to describe these as "created in V13 or later" and then on line 16, "created before V13". That is a more accurate way to describe these two different exports.

rwmcintosh: Reworded in 22be2f0.

- `Individuals and Expression Profiles created with version 13 or later` → PK-Sim snapshot + the values changed in MoBi
- `Individuals and Expression Profiles created before version 13` → PKML, verbatim

part-4/converting-v12-projects-to-v13.md
```diff
@@ -231,7 +231,7 @@ The goal is to arrive at a v13 model configuration whose every difference from t
...
 
-   A PK-Sim module created in v12 is a v12 artefact: it carries the v12 PBPK model structure and the v12 physiological and molecular database. Simply carrying it into a v13 project keeps all of that, and none of the v13 improvements are applied — the same reasoning as for [PK-Sim projects](../part-3/importing-exporting-project-data-models.md#exporting-project-to-snapshot--loading-project-from-snapshot). Each PK-Sim module must therefore be re-created in PK-Sim v13, re-imported into the MoBi project in place of the old one (**preserving each simulation's module order**), and **every** simulation that uses it rebuilt.
+   A PK-Sim module created in v12 is a v12 artefact: it carries the v12 PBPK model structure and the v12 physiological and molecular database. Simply carrying it into a v13 project keeps all of that, and none of the v13 improvements are applied — the same reasoning as for [PK-Sim projects](../part-3/pk-sim-snapshots.md). Each PK-Sim module must therefore be re-created in PK-Sim v13, re-imported into the MoBi project in place of the old one (**preserving each simulation's module order**), and **every** simulation that uses it rebuilt.
```
Yuri05, line 234: Would extend this part with:
_This is an **optional** conversion step, required only in case the user either wants the latest PBPK model structure and/or has the intention to export the MoBi project as snapshot (e.g. to use in qualification etc.)_

part-4/mobi-snapshots.md
```diff
@@ -0,0 +1,103 @@
...
+
+A project snapshot is a human-readable text file in [JSON format](https://en.wikipedia.org/wiki/JSON) that contains the information required to re-create the project. In contrast to a `*.mbp3` project file, which stores the fully built models, a snapshot stores the *inputs*: the modules and building blocks of the project, the configuration of each simulation, and the changes made by the user. When the snapshot is loaded, the simulations are rebuilt from that configuration.
+
+For a PBPK model built in PK-Sim®, the snapshot concept also carries the model *forward*: the PK-Sim® modules of a MoBi® project are stored as PK-Sim® snapshots and are re-created by a local PK-Sim® installation when the MoBi® snapshot is loaded, so they are rebuilt with the physiological and molecular database of the installed PK-Sim® version. See [Modularization concept](modularization-concept.md) and [Converting v12 projects to v13](converting-v12-projects-to-v13.md) for what this means for model migration.
```
Yuri05: ...the PK-Sim® modules, **Individuals and Expression Profiles** of a MoBi® project are stored ...

part-4/mobi-snapshots.md
```diff
@@ -0,0 +1,103 @@
...
+
+## What a MoBi project snapshot contains
+
+| Project content | How it is stored in the snapshot |
+| --- | --- |
+| PK-Sim® modules | as the PK-Sim® snapshot the module carries, embedded as a JSON sub-node. The module is **re-created by PK-Sim®** when the snapshot is loaded. |
+| Extension modules | as PKML, stored verbatim (Base64-encoded). They are restored as they were, not re-created. |
+| Individuals and Expression Profiles created with version 13 or later | as the PK-Sim® snapshot of the building block, plus the parameter values (and formulas) that were changed in MoBi® afterwards. The building block is re-created by PK-Sim® and the stored changes are re-applied on top. |
+| Individuals and Expression Profiles created before version 13 | as PKML, stored verbatim (Base64-encoded). |
```
Yuri05: Does the statement "_They are restored as they were, not re-created._" also apply here?

Yuri05 (approved):

